PR #1, opened 2018-04-10
RFC: Resources v2

vito:
[Rendered](https://github.com/vito/rfcs/blob/resources-v2/001-resources-v2/proposal.md)

Status: ~80% figured out. The remaining 20% is concerning performance/architectural implications that we should be sure to understand before going forward.

This is a very large proposal, heavily influenced by concourse/concourse#1707 and with a lot of prior planning in concourse/concourse#534. The discussion in 534 was growing stale, which was the catalyst for this new RFC format. Please leave feedback on individual lines instead of commenting directly so that the number of open discussion threads dwindles down as the proposal stabilizes and improves.

TODOs:

* [ ] https://github.com/concourse/rfcs/pull/1#discussion_r263907060


01-resources-v2/proposal.md
```diff
@@ -0,0 +1,285 @@
...
+* Add a `info` script which prints the resource's API version, e.g.
+  `{"version":"2.0"}`. This will start at `2.0`. If `/info` does not exist we'll
+  execute today's resource interface behavior.
+
+* Rather than running `/opt/resource/X`, discover the supported resource APIs by
+  invoking `info`. This allows us to be more flexible in what kinds of resources
+  we can support (versioned artifacts, notifications, ???), and where the
+  scripts may live (`/opt/resource` is very Linux specific).
+
+* Today's resource interface (`/in`, `/out`, `/check`) becomes more specifically
```
cwlbraa: something that was explicit in the original proposal was that "in" would become "get" and "out" would become "put", and feel like that was a solid rename choice that should carry forward to this proposal.

01-resources-v2/proposal.md
```diff
@@ -0,0 +1,285 @@
...
+  pitfall during resource development/debugging.
+
+
+## Introduction of "notifier" resource interface
+
+A resource may now implement a "notifier" interface. Notifications must be
+explicitly enabled in the pipeline. This is so that a resource type can't
+suddenly decide to include a notifier and start doing surprising things.
+
+A notifier has a single hook, `notify`, and receives some sort of JSON payload
```
cwlbraa: is this not already solved by put-only resources/artifacts? does it need to be a first class citizen?

01-resources-v2/proposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Is `destroy` general enough to be a part of the interface?
+
+It may be the case that most resources cannot easily support `destroy`. One
+example is the `git` resource. It doesn't really make sense to `destroy` a
+commit. Even if it did (`push -f`?), it's a kind of weird workflow to support
+out of the box.
+
+Could we instead just have `put` and ensure that we `check` in such a way that
+deleted versions are automatically noticed? What would the overhead of this be?
```
cwlbraa: the semantics for how to implement this in `check` sound difficult... does check now both get and return a list of a deleted resources? how am i supposed to figure out what those are, especially after they've been deleted?

01-resources-v2/proposal.md
```diff
@@ -0,0 +1,285 @@
...
+requiring that resource `check`s be side-effect free and able to run in parallel?
+
+There may be substantial security implications for this.
+
+
+## Is `destroy` general enough to be a part of the interface?
+
+It may be the case that most resources cannot easily support `destroy`. One
+example is the `git` resource. It doesn't really make sense to `destroy` a
+commit. Even if it did (`push -f`?), it's a kind of weird workflow to support
```
cwlbraa: for git, yes, it a weird workflow. does every resource need to implement all parts of the interface? Implementing noops is far less cumbersome for resource implementers than adding additional complexity to any one of the interfaces. If info could declare which bits of interface are getting implemented, and then the interaction between all of them could be described, that'd be even better.

01-resources-v2/proposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+Here are a few use cases that resources were sometimes used for inappropriately:
+
+## Single-state resources
+
+Resources that really only have a "current state", such as deployments. This is
+still "change over time", but the difference is that old versions become invalid
+as soon as there's a new one.
+
+These can now be done by always marking the old versions as "deleted".
```
cwlbraa: how can a resource implementer make this happen for every single put, while simultaneously supporting explicit deletes from the pipeline side? I'd like first class support for these sorts of resources (especially because there are so many of them) but after reading this, and thinking about this exact problem myself prior to reading this, I'm still uncertain as to what that might look like.

01-resources-v2/proposal.md
vito: True! I'm planning on that being part of the JSON schema returned by `/info` which I haven't gone into yet. Something like:

```json
{
  "version": "2.0",
  "interfaces": {
    "artifacts": {
      "get": "/opt/resource/get",
      "put": "/opt/resource/put",
      "check": "/opt/resource/check",
      "destroy": "/opt/resource/destroy",
      "spaces": "/opt/resource/spaces"
    }
  }
```

cwlbraa: oh nice, so you can also build a monolithic binary if you so choose! that'd be great.

01-resources-v2/proposal.md
vito: I wouldn't say it's "solved", from a design perspective. It's *doable*, but there are many drawbacks:

* Resources are meant to provide meaningful abstractions. If a whole class of them circumvent the majority of the interface, and act as glorified tasks instead, there's not much meaning to their use in a pipeline - they no longer describe a dependency lifecycle, they describe third-party hooks to execute in artbitrary places.
* "Partial" resource implementations makes it impossible to know what to expect from third-party resources. I don't think it's great to be limited in what verbs you can use, either because the author didn't have the time/motivation to implement the full thing or because it doesn't fit the interface well.
* Using an entire resource just for notifications results in a ton of clutter, both in the config and in the UI.

01-resources-v2/proposal.md
vito: Yeah, this part's fairly tricky. Here is one way I've thought of:

When `check` is given a version that's no longer available, it could indicate that it's gone by returning the latest version *plus* the version prior to it. Because the cursor version isn't returned, it would be marked as 'deleted', and any existing versions in the database that were previously *in between* the returned versions would be marked as deleted as well.

This would work for `git` because there's an inherent linearity to its history wherein its commits point directly to a parent commit, but I haven't reasoned about it for other implementations or thought about other more complicated situations.

01-resources-v2/proposal.md
cwlbraa: I see where you're coming from. I started rebutting by writing this out as interface composition but when I did the difference between the two was immediately apparent: a notifier doesn't emit versions.

01-resources-v2/proposal.md
vito: This hinges on `check` being able to detect that the version has gone away. If that's doable (big if, but I think it might be), then `put` can just be a general side-effecter, and `check` is the verification that things were either deleted (by noticing only the new version, and that the cursor version given to it is no longer there) or created (by having new versions).

01-resources-v2/proposal.md
cwlbraa: even in the git case, what if I delete HEAD?

01-resources-v2/proposal.md
vito: Yeah sorry, the Elm mainly helps me right now - it's easier to move around which types goes into which function. Once things gel a bit the doc will get more detailed/accurate. :)

01-resources-v2/proposal.md
cwlbraa: > I think it's a smell if there are parts of a general interface that are impossible to satisfy with a substantial portion of the implementations

I totally agree and I think we're probably smelling the same inadequacy. if it didn't sound super absurd on its face, I'd think that maybe what we need is something like an antiartifact-- 

it's got `anticheck` and `delete`, but obviously no get. both emit deleted versions in chronological order?

01-resources-v2/proposal.md
vito: In the git case, if you delete `HEAD` (assuming you mean `push -f` of the commit prior to it), the version that was *previously* at `HEAD` would be given to `check` (since all checks run from a cursor, except for the initial one).

Today's resource interface says that if the given version is no longer there, it shouldn't be returned - instead, the new "latest" version should be returned.

So, assuming you started from this history:

```
commit 1
commit 2
commit 3
commit 4 <- HEAD
```

If you were to `push -f` this new history:

```
commit 1
commit 2
commit 3.1
commit 4.1
```

...then today, the the `check` script would return:

```
commit 4.1
```

And we could thus know that at least `commit 4` is no longer with us.

That leaves `commit 3` still there, though, so we haven't achieved *full* detection yet. This is where I'm stuck. Just thinking out loud, maybe the trick is to be able to `check` for ancestry in order to close the gap? Then the ATC could walk backwards in the history until either all prior versions are wiped out or until the loop is closed (by finding commit 4.1 -> commit 3.1 -> commit 2, which already exists, so the job would be done).

01-resources-v2/proposal.md
vito: Yep! We'll want to be careful not to tie ourselves to shell command line parsing, though, so this might take a form similar to the task's [`run` block](https://concourse-ci.org/tasks.html#task-run) instead.

01-resources-v2/proposal.md
vito: Funny that you mention `anticheck` - sounds similar to the ancestry check I just commented here: https://github.com/concourse/rfcs/pull/1#discussion_r182265596

01-resources-v2/proposal.md
cwlbraa: I know the resource that I maintain is simultaneously "single-state" and probably also spatial-- ie there's multiple underlying states, but each one invalidates itself when you update it. At any given time it's easy for me to list all the valid versions, but linearizing across spaces is cumbersome. 

When you say in an above comment
> When check is given a version that's no longer available, ...
If this is happening on a per-space basis, I think check not returning the cursor version makes enough sense-- things are pretty easy to linearize. My head swims a bit if think about deleting "old" versions or deleting versions from a big stack, though. Does "check" receive every version that's been emitted from `put`? does it receive every version that's been deleted?

01-resources-v2/proposal.md
cwlbraa: Your example is slightly more difficult and more illustrative than what i meant initially... 

I think I meant if I `push -f` over your 4 commit example:
```
commit 1
commit 2
commit 3 <- HEAD
```
then check would get `commit 4` and it'd return `commit 2; commit 3`?

vito:
To be honest I wrote that it could include the prior version earlier but I
don't think it helps enough. I think the ancestral scan would be necessary
instead. So in this case it would just return `commit 3` and Concourse
would know that the given version is deleted.

01-resources-v2/proposal.md
vito: (re-posting 'cause email reply doesn't keep context)

To be honest I wrote that it could include the prior version earlier but I don't think it helps enough. I think the ancestral scan would be necessary instead.

So in this case it would just return `commit 3` and Concourse would know that the given version is deleted (since it wasn't returned).

01-resources-v2/proposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+* Remove the distinction between `source` and `params`; resources will receive a
+  single `config`. The distinction will remain in the pipeline. This makes it
+  easier to implement a resource without planning ahead for interesting dynamic
+  vs. static usage patterns, and will get more powerful with #684.
+
+
+## Changes to Versioned Artifact resources
+
+* Add a `spaces` actopm, which is used to discover spaces.
```
dprotaso: action? your right hand shifted left one key

vito: Haha, I'll fix it.

01-resources-v2/proposal.md
itsdalmo: Couple of questions on this:
- Notifiers will still be visualised in the pipeline so that it's apparent if it has failed?
- Are notifiers only fed a JSON describing just the events they are "subscribed" to? In your example (`01-resources-v2/notifications.yml`), it seems like the slack notification would trigger and send a message if either the whole pipeline succeeded, or any part of it failed?

If I've understood things correctly in my last question, and you can't use notifier resources in `on_success`/`on_failure` as we do today, I have a few pipelines in production now that would not be able to replace the current slack notification resource with the new "notifier" version. Because, the pipelines aggregate information from multiple "get" steps, piece it together in one task, and output it to slack as the last "put" step.

01-resources-v2/branch-gen.yml
```diff
@@ -0,0 +1,39 @@
...
+    config:
+      platform: ...
+      image_resource: ...
+      outputs:
+      - name: generated-repo
+        # has the generated code committed to the repo
+      - name: branch-name
+        # has a 'name' file with `gen-(some deterministic hash)`
+  - put: atc-gen
+    params: {branch_name: branch-name/name, repository: gen-repo}
```
itsdalmo: Is `repository: gen-repo` supposed to be `atc-gen`?

01-resources-v2/branch-gen.yml
```diff
@@ -0,0 +1,39 @@
+resources:
+- name: atc
+  type: git
+  source: {uri: https://github.com/concourse/atc}
+  space: {branch: master}
```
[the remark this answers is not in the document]
itsdalmo: Thinking about build matrices, wouldn't this need to be a map of arrays (i.e.`map[string][]string` in go)? E.g.:

```
spaces:
  branch: [master, development]
```

Or, is it called `space` on a resource, and `spaces` on a task? If so, why?

01-resources-v2/commit-status.yml
```diff
@@ -0,0 +1,30 @@
...
+  source:
+    repository: concourse/atc
+    access_token: ((token))
+
+jobs:
+- name: atc-pr-unit
+  plan:
+  - get: atc-pr
+    trigger: true
+    spaces: all
```
itsdalmo: What happens to `versions: every`? With spaces, would it be possible to build every single commit on all open pull requests by setting `spaces: all` and `versions: every`?

01-resources-v2/proposal.md
itsdalmo: In the resources I've written, the first `check` usually has to fetch multiple "potential" versions, figure out which is the last, return that (last) version and discard the rest. Would it make sense that `check` returns the complete list of _all_  (or 100-1000) versions on the first check (i.e. when it's passed an empty version)? So if we notice that a version is missing like in the scenario above, it just resets and start over like it was the first check again?

Reason I'm bringing this up is that it sounds like the "ancestral scan" would involve a lot of API calls to iterate backwards if someone decided to rebase `concourse/concourse` with random squashes over 8000 commits.

If the goal is to keep a complete historical record of all versions that passed through Concourse, this will obviously not do. But the way I see it, someone force pushing over a commit and us removing that version, does not change the fact that one or more artifacts might have been built on that particular revision of the code - so the get/put/build logs would still reference a version which no longer exists, which means they are our "source of truth".

01-resources-v2/proposal.md
itsdalmo: Spaces seem to be tied to stuff that can be discovered by resources (e.g. branches), but I'm not sure I see anything in the RFC in terms of build matrices - at least not the type that was in the original proposal for spaces, such as specifying the ruby version?

Also, would it be possible for a resource to support multiple `spaces`? For instance, the `git` and `github-pullrequest` resources already have `path` and `ignore_paths` parameters. Could `path` theoretically be a space on these resources as well?

Example:
```yml
resources:
- name: atc
  type: git
  source: {uri: https://github.com/concourse/atc}
  spaces:
    branch: [master]
    path: [concourse/atc/*, concourse/fly/* ]
```

So a commit to master which changes both paths would produce two versions, each with a different path in their `spaces` payload. The reason I'm asking about this is because it would allow us to have spaces for monolithic repositories. Would it would be considered an anti-pattern to do it this way, since what would be pulled in `get` would be the complete repository? Also, would this put too much responsibility on the resources in order to implement all the "spaces" required to make build matrices?

01-resources-v2/branch-gen.yml
vito: Nope, it should be `generated-repo` (the task output). Fixing.

01-resources-v2/branch-gen.yml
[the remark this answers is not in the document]
vito: On a resource, you would define a single `space: ...` if you only ever care to deal with a single space everywhere by default (similar to today's behavior, only the space-related config would be under `space:`, not `source:`).

It's singular because at least when we were initially planning this it felt like having resources configure multiple spaces would be harder to keep track of vs. having each job define its spaces on a case-by-case basis.

01-resources-v2/commit-status.yml
vito: Yep, that's how the two would interact.

01-resources-v2/proposal.md
vito: Correct - spaces would only be useful for build matrixes when the things you're matrixing over can be represented as resources. Our theory/hope is that this should cover most things. For example, matrixing over Ruby versions would be done by using the Ruby docker images, which are semver'd:

```yaml
resources:
- name: ruby
  type: docker-image
  source: {repository: ruby}

jobs:
- name: unit
  plan:
  - get: ruby
    spaces: [{tag: 1.*}, {tag: 2.*}]
  - get: my-repo
  - task: unit
    file: my-repo/ci/unit.yml
    image: ruby
```

We don't currently have a plan for matrixes that are just static config (e.g. `params` to tasks). Spaces is a resource-only feature right now, and it may be unwise to stretch its scope any further at this point. I suspect this may end up leading to more manageable patterns anyway - by forcing everything into resources you'll be working with checked-in code/config rather than reconfiguring pipelines.

As long as it's easy to implement custom resource types, we could see specialized resource types for supporting workflows like giant monorepos and path-based spaces. I see the future tending towards many workflow-specific resource types vs. one big swiss-army `git` resource type.

01-resources-v2/proposal.md
itsdalmo: Thanks for the information! Personally Regarding a new `notify:` step type, I think it's a good idea to try to keep the number of keywords to a minimum (as far as possible). But perhaps what I'd suggest is that we skip this:

```
A notifier has a single hook, `notify`[...]
```

What if notify is just another interface that can be implemented by resources, which then makes them eligible for the `notify:` configuration (as you have drafted in your example). But if it also implements `put:`, you can continue to use it as is currently done in pipelines (with custom/aggregated text)?

Perhaps I don't have the correct vision, but impression from the resources I've used (and some I've written) is that only implementing e.g. `check` and `get`, but not `put`, and the other way around, are valid use cases and "working as intended". If `check` and/or `get` are no-op, then I think it would make sense to drop the requirement that resources return versions also. 

Since resources v2 will return `info` with all the interfaces they support, my thinking was that the list of interfaces implemented by the resource could determine how it's represented in Concourse. So e.g. the `slack-notification-resource` could implement only `info, `notify` and `put`, and therefor not return any version, only references to builds and steps that triggered a `notify`, and the metadata emitted at the time?

01-resources-v2/proposal.md
```diff
@@ -0,0 +1,285 @@
...
+
+-- data on disk
+type alias Bits = ()
+```
+
+## Versioned Artifacts interface
+
+```elm
+spaces  : Config -> Set Space
+check   : Config -> Space -> Maybe Version -> List (Version, Metadata)
```
itsdalmo:
```
Spaces are checked independently and have their own version history, 
so this should work fine. You would represent each state as its own space, 
if they can be independently updated.
```
From the elm code and your comment above I understand this as saying that `check` will be passed a specific space to check for new versions? So, for the `git-resource` Concourse would:

1. Ask it for a list of `spaces`, which would then return all the branches in the repository.
2. For each space returned by `spaces`, it would call `check` for that specific space.

This means that the `git-resource` could never benefit from using the Github V4 API (GraphQL), which could return the last commit for all branches in a single API call - since it would essentially be forced to first list all branches, and then iterate over them to find the last commit.

If I understood it correctly, it does not sound ideal?

vito: @itsdalmo Interesting point. (Splitting hairs, but I wouldn't expect `git-resource` to hit the GitHub API, since it's generic `git` and not `github`, but your point still stands since there could be some GitHub-specific resource type that does.)

I went with one-check-per-space as it seemed like it would result in relatively simple code for `check`, and be an easier transition from today's resource implementations.

But having a single request is worth considering, with the caveat that having everything in one request could mean less parallelism, since it all has to happen in one container with shared state.

Taking a look at today's `git` resource, which works by cloning the repo and running `git log`, it could actually work this way pretty easily by running `git log <branch>` for each space. This should even support `paths`/`ignore_paths`.

One big win would also be fewer containers running. Folks already complain about the number of `check` containers, and the current design could make that worse by having an additional container for each space.

So, I'll toy with this idea! Thanks!

01-resources-v2/proposal.md
vito: @itsdalmo Slight clarification: we would never remove versions, we would mark them deleted, and probably have them shown separately from the "source of truth" version history. This is purely to preserve those get/put associations. These 'soft deleted' versions wouldn't be available as inputs, they'd just be archived.

I like the simplicity of having resources just say "ok screw it here's everything I know" if the given version is gone, and diffing the history to notice which versions are gone. We'd have to be careful to design an API that's hard for resource authors to mess up, though. Everyone screws up pagination the first time. ;)

01-resources-v2/proposal.md
itsdalmo: Great! I mentioned the `git-resource` but I what I was really thinking about was a `github-pr-resource` I've written myself, which uses the GraphQL api for Github and would reap 0 benefits for doing so if `check` was passed one space to check - so single request gets a +1 from me at least 😄

01-resources-v2/branch-gen.yml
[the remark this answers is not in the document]
itsdalmo: Yep, makes sense to me!

01-resources-v2/proposal.md
vito: @itsdalmo I prefer to not overload verbs when possible, especially if they would have a different "function signature" (e.g. `put` for notifier doesn't return versions). I think today's pattern of partial resource implementations has weakened the strength of the abstraction and achieved a local maxima. It makes it harder to reason about resources as meaningful objects - they become just bags of verbs.

So I'd rather take this opportunity to introduce more meaningful concepts (i.e. notifications/status reports) rather than overload the verbs already being used for versioned artifacts. Having more verbs which always mean one thing, and are always tied to a meaningful interface, will be cleaner in the long run, even though mechanically they may be fairly similar.

With the addition of `/info` you can then have a single image that supports both the notifier interface and the versioned artifact interface. This would be useful for things like pull request resources, which would use the notifier interface to reflect the build status back to GitHub, while using the versioned artifact interface for the actual brunt of the work (finding PRs/commits and fetching them).

01-resources-v2/proposal.md
vito: @cwlbraa Just circling back: I think this is now resolved. I've changed `put` to return two sets: a set of created versions, and a set of deleted versions, per-space. There was really no way `check` could magically discover deleted versions in all cases (only `git`). Tangentially related, I've changed `check` to return all versions in the event that given version is no longer found, per @itsdalmo's suggestion.

01-resources-v2/git-example/artifact
```diff
@@ -0,0 +1,104 @@
...
+      ]
+    }
+  end
+end
+
+case ARGV[0]
+when "check"
+  $request = {
+    config: {uri: "https://github.com/vito/booklit"},
+    from: {master: {ref: "40bc6986197e411471d306bb8eb3a21c5b5f9d26"}}
```
itsdalmo: Bit confused here. From [this](https://github.com/vito/rfcs/blame/217f5754433de8a702e46c1fb8a90621bb6c7816/01-resources-v2/proposal.md#L54) I understand that `check` should receive a `Space` and a `Version`. And [`Version`](https://github.com/vito/rfcs/blame/217f5754433de8a702e46c1fb8a90621bb6c7816/01-resources-v2/proposal.md#L42) should be a dict ( e.g. `{ "ref": "" }`), while [`Space`](https://github.com/vito/rfcs/blame/217f5754433de8a702e46c1fb8a90621bb6c7816/01-resources-v2/proposal.md#L39) should just be a string?

Based on the above, I'd expect to see something like this:

```json
{
  "space": "master",
  "version": { "ref": "40bc6986197e411471d306bb8eb3a21c5b5f9d26" }
}
```

01-resources-v2/git-example/artifact
```diff
@@ -0,0 +1,104 @@
...
+    }
+  end
+
+  response = JSON.dump(spaces)
+  puts response
+  puts response.size
+when "get"
+  $request = {
+    config: {uri: "https://github.com/vito/booklit"},
+    space: {branch: "master"},
```
itsdalmo: Ah, here is the payload I was expecting 👍

01-resources-v2/proposal.md
```diff
@@ -51,11 +51,9 @@ type alias Bits = ()
 ## Versioned Artifacts interface
 
 ```elm
-spaces  : Config -> Set Space
-check   : Config -> Space -> Maybe Version -> List (Version, Metadata)
+check   : Config -> Dict Space Version -> Dict Space (List (Version, Metadata))
 get     : Config -> Space -> Version -> Bits
-put     : Config -> Bits -> (Space, Set Version)
-destroy : Config -> Bits -> Dict Space (Set Version)
+put     : Config -> Bits -> Dict Space { created : Set Version, deleted : Set Version }
```
itsdalmo: A thought that occurred to me here, is that in the context of a `git` resource I think it is more likely that a `get` (with `version: every`) will discover that a version is missing after force push, than `check` or `put` (which will not support deleting refs anyway)?

Might it be an idea to allow get to signal that the version being fetched no longer exists, rather than just exit with an error (if someone has force pushed pushed after the build was started)?

itsdalmo: If it makes sense to allow `get` to signal if a version is missing, I'm wondering if it also would make sense (since `get` would not have deleted anything itself) to just rename the keys for the returned objects to avoid mistakes on the part of the resource implementors:

- `versions` and `created` renamed to `added`.
- `deleted` renamed to `removed`

So `check` would return: 
```json
{
  "space": "master", 
  "added": [{ 
    "ref": "sha", 
    "metadata": <map>
  }]
}
```
And `get` _could_ return:
```json
{
  "space": "master", 
  "removed": [{ 
    "ref": "sha"
  }]
}
```
While `put` would return (removed optional):
```json
{
  "space": "master", 
  "added": [{ 
    "ref": "sha"
  }],
  "removed": [{ 
    "ref": "sha"
  }]
}
```

Might need better naming, but I think it might be the case that `get` will be better at discovering missing versions in some cases. E.g., with the git resource on `version: every` there might already be a later version than the one currently being passed to `get`, in which case I'd like something to let `check` know that we need to start fresh?

01-resources-v2/git-example/artifact
vito: Woops that should actually be `space: "master"` - I fixed it in a later commit. I switched from JSON objects to simple strings since it seemed more natural to use them as keys. I'm slightly iffy on that decision still but it does simplify implementation (and probably UI/yaml).

itsdalmo: One thing that might be worth noting; by using a simple string we lose the ability to produce (easily parsable) spaces that are defined by more than one piece of information. E.g., with the current resource interfaces, I defined a version by the Pull Request ID and Commit ID in this resource: [github-pr-resource](https://github.com/itsdalmo/github-pr-resource/blob/master/models.go#L53-L57).

Not sure if this is something that would be considered an anti-pattern, or whether there will ever exist spaces that can only be defined by more than one piece of information?

01-resources-v2/proposal.md
vito: Hmm, I like the symmetry here but I could see a few problems with it.

For Git repos at least, the `check` should notice a force-push pretty quickly because the latest version's `ref` should no longer be valid.

There could also be a tricky scenario where an older version of a resource that the build is trying to fetch is temporarily unavailable. Marking it 'deleted' would prevent another build from attempting to fetch it. I suppose `check` could re-discover it, but only if it's the latest version.

01-resources-v2/git-example/artifact
vito: I took a liberty here and used the space as the key in the map, instead of having a list of JSON objects like that. It makes it easier to find the 'from' version when checking for each branch.

This is really the only reason I switched from JSON objects to strings for the space representation.

01-resources-v2/git-example/artifact
vito: So far I haven't seen a situation where you'd need multiple pieces of information for *spaces*, but that's worth considering. Although it could get ugly quickly in pipeline .yml since space specifications would end up pretty verbose (`spaces: [{foo: v1.*, bar: baz}]`).

For the PR example the space should just be the PR number, and commits would be the versions within the space (which can now just have a single key).

01-resources-v2/git-example/artifact
itsdalmo: Correct me if I'm wrong - but wouldn't Go code look like this then:

```go
type CheckRequest struct {
  config  Config `json:"config"`
  version map[string]map[string]string `json:"from"`
}
```

If so, that seems less "clean" than having:
```go
type CheckRequest struct {
  config  Config `json:"config"`
  version Version `json:"version"`
}

type Version struct {
  Space string `json:"space"` // Or map[string]string
  Version map[string]string `json:"version"`
}
```

vito: To be fair, you've got a `map[string]string` there where really the resource would have its own type. So it'd be:

```go
type Version struct {
  Ref string `json:"ref"`
}

type CheckRequest struct {
  Config Config `json:"config"`
  From map[string]Version `json:"from"`
}
```

It may not be as pretty on the type definition, but it's definitely prettier when it comes to finding the version to check from:

```go
for _, branch := range branches {
  var version Version
  for _, f := range request.From {
    if f.Space == branch {
      version = f.Version
    }
  }

  // git log {version.Ref}..HEAD
}
```

vs.

```go
for _, branch := range branches {
  version := request.From[branch]
  // git log {version.Ref}..HEAD
}
```

Of course, you could clean it up with a helper method, but it's still more code that the author has to write.

itsdalmo: I see what you mean, and it does look better:) Just have one question; `check` requests will contain the latest version for all discovered spaces? (And not just the latest across spaces?). In your Ruby example it's just a single space and version, figure it might be worth updating the example to reflect what will actually be passed to the resource.

vito: Yep. I'll update the example to be more clear.

vito:
I'm going to cut the scope of this RFC down a bit by removing all the stuff concerned with notifications. I think there's a lot more to explore there and it warrants its own RFC. I'd rather wrap up the new artifacts interface sooner so that we can get support for spaces out the door.

I also just realized we haven't at all discussed the idea of schema validation. I'm going to cut that out, too - we should be able to add that in later, again as a separate RFC.

01-resources-v2/git-example/artifact
```diff
@@ -0,0 +1,141 @@
...
+    end
+
+    space_name = b.name.sub("#{b.remote_name}/", "")
+
+    has_latest = false
+    commits = []
+
+    walker = Rugged::Walker.new(repo)
+    walker.sorting(Rugged::SORT_TOPO|Rugged::SORT_REVERSE)
+    walker.simplify_first_parent
```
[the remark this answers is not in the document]
vito: This makes checking the Linux repo a walk in the park:

```
~/w/r/0/git-example (resources-v2) $ time -f "%P cpu %M ram in %E %I/%O" ./collect-all-versions 
master walk: 0.9525529199745506s
master commits: 10000 (latest: false)
dump: 0.16487152903573588s
checked master: 1da177e4c3f41524e886b7f1b8a0c1fc7321cac2..36920e069a87c6fccffbc0ec5e74985d94d3bcc4
master walk: 0.8437563210027292s
master commits: 10000 (latest: false)
dump: 0.18869470001664013s
checked master: 36920e069a87c6fccffbc0ec5e74985d94d3bcc4..7d2284b09fd967b5cad6f03c6d4a5c064f37b855
master walk: 0.850168353994377s
master commits: 10000 (latest: false)
dump: 0.22101561503950506s
checked master: 7d2284b09fd967b5cad6f03c6d4a5c064f37b855..d1e98f429aa10132b3010ba3b0be47552a2eb14b
master walk: 0.7699124459759332s
master commits: 10000 (latest: false)
dump: 0.19795792998047546s
checked master: d1e98f429aa10132b3010ba3b0be47552a2eb14b..b0365c8d0cb6e79eb5f21418ae61ab511f31b575
master walk: 0.7383112910320051s
master commits: 10000 (latest: false)
dump: 0.2800299650407396s
checked master: b0365c8d0cb6e79eb5f21418ae61ab511f31b575..559b6d90a0beb375c46dffe18133012bfa29f441
master walk: 0.35654218099080026s
master commits: 4940 (latest: true)
proc-cmdline walk: 0.7496353160240687s
proc-cmdline commits: 5060 (latest: false)
dump: 0.23467062803683802s
checked master: 559b6d90a0beb375c46dffe18133012bfa29f441..894b8c000ae6c106cc434ee0000dfb77df8bd9db
checked proc-cmdline: 1da177e4c3f41524e886b7f1b8a0c1fc7321cac2..c7306c02874bf4d22bc0b1dbea34282d0b9a3df1
master commits: 1 (latest: true)
proc-cmdline walk: 0.922503334004432s
proc-cmdline commits: 9999 (latest: false)
dump: 0.21390072000212967s
checked master: 894b8c000ae6c106cc434ee0000dfb77df8bd9db
checked proc-cmdline: c7306c02874bf4d22bc0b1dbea34282d0b9a3df1..193faea9280a809cc30e81d7e503e01b1d7b7042
master commits: 1 (latest: true)
proc-cmdline walk: 0.8317700829938985s
proc-cmdline commits: 9999 (latest: false)
dump: 0.18607709102798253s
checked master: 894b8c000ae6c106cc434ee0000dfb77df8bd9db
checked proc-cmdline: 193faea9280a809cc30e81d7e503e01b1d7b7042..d91958815d214ea365b98cbff6215383897edcb6
master commits: 1 (latest: true)
proc-cmdline walk: 0.8469380890019238s
proc-cmdline commits: 9999 (latest: false)
dump: 0.19501436594873667s
checked master: 894b8c000ae6c106cc434ee0000dfb77df8bd9db
checked proc-cmdline: d91958815d214ea365b98cbff6215383897edcb6..568133ebda39f7507759a744fa9cf4d5097bad2f
master commits: 1 (latest: true)
proc-cmdline walk: 0.7004600089858286s
proc-cmdline commits: 9999 (latest: false)
dump: 0.21844543900806457s
checked master: 894b8c000ae6c106cc434ee0000dfb77df8bd9db
checked proc-cmdline: 568133ebda39f7507759a744fa9cf4d5097bad2f..45f7fdc2ffb9d5af4dab593843e89da70d1259e3
master commits: 1 (latest: true)
proc-cmdline walk: 0.6632431410253048s
proc-cmdline commits: 9594 (latest: true)
dump: 0.3072412370238453s
checked master: 894b8c000ae6c106cc434ee0000dfb77df8bd9db
checked proc-cmdline: 45f7fdc2ffb9d5af4dab593843e89da70d1259e3..5ab8271899658042fabc5ae7e6a99066a210bc0e
master commits: 1 (latest: true)
proc-cmdline commits: 1 (latest: true)
dump: 0.00010683003347367048s
checked master: 894b8c000ae6c106cc434ee0000dfb77df8bd9db
checked proc-cmdline: 5ab8271899658042fabc5ae7e6a99066a210bc0e
time for stable check: 0.49638886s
total time for all checks: 20.219301953s
total versions: {"master":54935,"proc-cmdline":54645}
89% cpu 584408 ram in 0:20.34 0/96
~/w/r/0/git-example (resources-v2) $ 
```

Peak memory usage of ~584MB, only took 20 seconds, and reduced the collected versions to ~55k down from the ~766k without `--first-parent`:

```
~/w/r/0/g/linux ((894b8c00…)) $ git log --oneline | wc -l
766654
~/w/r/0/g/linux ((894b8c00…)) $ git log --oneline --first-parent | wc -l
54935
```

01-resources-v2/proposal.md
```diff
@@ -0,0 +1,412 @@
...
+type CheckRequest struct {
+  Config Config            `json:"config"`
+  From   map[Space]Version `json:"from"`
+}
+```
+
+The first call will have an empty object as `from`.
+
+Any spaces discovered by the resource but not present in `from` should collect
+versions from the beginning.
```
vito: I'm still going through the performance implications of collecting *every* version for *every* space for *every* resource in *every* pipeline. This is what I'm most cautious about.

## Disk usage on DB

Collecting all the metadata for every version impacts disk usage on the database. I mean...that's just a fact, but I don't think it's a dealbreaker. This usage should be dwarfed by the space consumed by build logs, both in volume and rate of increase.

Mitigation: we may be able to share data between teams/pipelines by associating the versions/metadata to the global "resource cache" objects rather than per-pipeline entities. Even if we don't, I still think this shouldn't be a huge deal.


## Time/CPU/RAM consumption on workers

How do we emit the versions and metadata for, say, 700k commits to a monolithic git repo? How often will resource authors mistakenly implement this by holding everything in RAM?

Mitigation: I've added `has_latest: true` to the response, which will be required for Concourse to consider the versions to be "fully collected" and start using the versions for pipeline scheduling. This makes batching the versions/metadata held in RAM fairly easy (checking always goes chronologically and can be done with arbitrary batch sizes at the resource type author's discretion). Downside: it will take time to collect everything on first pipeline configuration. We can check spaces that don't have their latest version more aggressively until everything is collected.

## Scheduling performance

The current approach to pipeline scheduling works by loading literally every version of every resource into memory. This at least only loads database IDs (no version strings and certainly no metadata), but the sheer volume of versions to walk through and try when finding input candidates could really clog things up.

Mitigation: we could modify the scheduling algorithm to only load the latest 1000 or so versions of each resource at a time, and lazy-load older ones as needed (if the given set yields no input candidates, load 1000 more of everything). This should speed up scheduling efficiency in most cases anyway (it should be very infrequent that the latest valid candidate is a very old version), and reduce the memory usage on the `web` node overall (because the versions db is cached per-pipeline).

@itsdalmo Any thoughts/concerns on this?

itsdalmo: Hey!

> How often will resource authors mistakenly implement this by holding everything in RAM?

I think I'd be guilty of holding everything in memory myself, but from your tests it seems like this would be fine for everything but the edge cases (which I consider the Linux repo to be). And even with the linux repo the absolute memory usage is not that crazy, but perhaps the memory usage relative to other resources might trip people up?

> Mitigation: I've added has_latest: true to the response, which will be required for Concourse to consider the versions to be "fully collected" and start using the versions for pipeline scheduling.

So, I can e.g. paginate an API and emit versions, and only when I get to the last page and emit `has_latest: true` will Concourse schedule a task which has `trigger: true`? Is this only for the initial check, or for all checks? If so, how does it work with `every: true` (and/or ignore it on the initial check)?

Overall I think this is a good mitigation which makes it a lot easier to write lightweight resources. One way of _ensuring_ that resources are written in a lightweight manner might be to say that each version should be written individually to stdout, so resource authors would be encouraged to emit versions straight from a for-loop, rather than growing a temporary array.

What would be the drawbacks of doing that? We'd need to include the `space` for every version (rather than every array), and all the flushes to stdout might degrade performance?

> Mitigation: we could modify the scheduling algorithm to only load the latest 1000 or so versions of each resource at a time, and lazy-load older ones as needed (if the given set yields no input candidates, load 1000 more of everything). This should speed up scheduling efficiency in most cases anyway (it should be very infrequent that the latest valid candidate is a very old version), and reduce the memory usage on the web node overall (because the versions db is cached per-pipeline).

Perhaps I'm misunderstanding this, but in relation to `every: true` - it sounds like  it would either have to continuously loop over all versions to ensure that none have been missed, or that it would not be guaranteed to build every version? If the latter is the case, then I think the behaviour needs to be documented, or ideally that `every: true` should instead be `every: N` (where N could be 1000) or something to that effect?

vito: > And even with the linux repo the absolute memory usage is not that crazy, but perhaps the memory usage relative to other resources might trip people up?

It can actually get pretty high because we're returning all the metadata now (which includes commit messages). I believe it got up to a few gigabytes in my testing. The windowing, plus the first parent walking simplification, helps dramatically.

> So, I can e.g. paginate an API and emit versions, and only when I get to the last page and emit has_latest: true will Concourse schedule a task which has trigger: true? Is this only for the initial check, or for all checks? If so, how does it work with every: true (and/or ignore it on the initial check)?

I think it applies to all checks. It'll apply to all jobs, and it shouldn't matter whether `trigger: true` is set because that just influences whether Concourse auto-queues a build, not how the candidates are determined.

Here's how I think it'd work:

1. Run `check`, get `has_latest: false`. Save the versions, with a `partial` column set to `true` in the database. (This is a new column I just made up that could maybe use a better name.)
1. Modify the scheduler to only load versions with `partial` as false.
1. Run `check` again, this time it gets `has_latest: true`. Save the versions, and then set `partial` to `false` for all versions of the space.
1. Next time the scheduler runs, the full batch of versions will now be available.

This logic applies to both the initial check (because it may take a while to collect everything) and any subsequent checks (because a ton of versions may appear at once, and people would be annoyed if we started running from some random commit in the middle).

Also this should have no impact on `version: every` (assuming that's what you mean). It's just to prevent windows of time where we have partially saved data that would otherwise result in running things the user doesn't want.

> Overall I think this is a good mitigation which makes it a lot easier to write lightweight resources. One way of ensuring that resources are written in a lightweight manner might be to say that each version should be written individually to stdout, so resource authors would be encouraged to emit versions straight from a for-loop, rather than growing a temporary array.
> 
> What would be the drawbacks of doing that? We'd need to include the space for every version (rather than every array), and all the flushes to stdout might degrade performance?

I considered this, and it's really tempting, but I don't quite trust Garden's process output streaming enough. If the process emits enough output so quickly that the consumer can't keep up, Garden will begin dropping chunks of output so that the server doesn't get backpressured. Come to think of it, that may be reason enough to not use `stdout` for the response at all, now that it's going to be much larger. We might have to change the interface so that `check` writes the response to a file. (This would at least match the behavior of `put`.)

@topherbullock Would writing the resource's response to a file in the container make k8s support (concourse/rfcs#2) any easier, btw? (As opposed to streaming responses on `stdout`.)

> Perhaps I'm misunderstanding this, but in relation to every: true - it sounds like it would either have to continuously loop over all versions to ensure that none have been missed, or that it would not be guaranteed to build every version? If the latter is the case, then I think the behaviour needs to be documented, and perhaps something users can tweak on the resource configuration?

Good catch, I haven't thought about how that would work. `version: every` is currently implemented by [checking if the prior version has been used](https://github.com/concourse/atc/blob/9b578913f7f273197efd0fb1866558da88d4e74d/db/algorithm/input_candidates.go#L24-L47) - we would have to ensure this doesn't get tripped up by the windowing on the `.Peek` call.

jchesterpivotal:
I'm relieved to see the scope-decomposition.

For the version-explosion problem, do we have a good pathological example?

01-resources-v2/proposal.md
vito: Follow-up: writing the response to a file or stdout doesn't make it any easier or harder for k8s. I went with writing to a file for consistency with `put` and safety (since we can't 100% trust the streaming process output).

01-resources-v2/proposal.md
```diff
@@ -0,0 +1,413 @@
...
+default. If there is no default space, the user must specify it explicitly in
+the pipeline, either by configuring one on the resource (`space: foo`) or on the
+`get` step (`spaces: [foo]`).
+
+The command should then emit the collected versions and default space (if any)
+as following JSON response structure the file specified by `response_path`:
+
+```go
+// CheckResponse returns the detected spaces and the default space, if any.
+type CheckResponse struct {
```
vito: One last thing I'm considering before closing out this proposal: should we write one big JSON response body, or have a more stream-based response interface?

For example, the response could look something like:

```js
{"spaces":["a","b","c"],"default_space":"a"}
{"space":"a","version":{"v":"1"},"metadata":[]}
{"space":"a","version":{"v":"2"},"metadata":[]}
{"space":"a","version":{"v":"3"},"metadata":[]}
{"space":"b","version":{"v":"1"},"metadata":[]}
{"space":"b","version":{"v":"2"},"metadata":[]}
{"space":"b","version":{"v":"3"},"metadata":[]}
// ...
```

So there'd be an initial JSON object returning the spaces and default space, and from then on it'd be a stream of versions associated to each space. This would enable resource authors to write versions as they find them rather than hold them all in memory and do one big JSON marshal. With this design could get rid of `has_latest` - disk space would now be the "at risk" resource, but it should be a lot cheaper than RAM.

The interface itself is a bit more complicated but I think it'd actually make it harder for resource authors to make mistakes. Authors would notice a broken implementation of a more complicated interface early, rather than implement the easier interface but then have emergent behavior like using a ton of RAM on large repos.

@itsdalmo @cwlbraa @topherbullock whaddaya think?

itsdalmo: Sorry for the late reply!

I'm wondering if it would be possible to have Concourse discover the possible spaces from the set of spaces as it reads this records? E.g., in the example below it would read one record at a time and keep a set in memory which ends up as (`["a", "b"]`):

```json
{"space":"a","version":{"v":"1"},"metadata":[]}
{"space":"b","version":{"v":"1"},"metadata":[]}
```

The reason I ask is because in e.g. a `github-pullrequest-resource` which uses the GraphQL API, it's possible (and desirable because it saves API calls) to just list all open PR's directly, irrespective of which branch or base it has. So in that case, the resource itself could not write the complete set of spaces (the first record in your example) until it had retrieved every open PR from the API, stored them in memory and then looped over them to create a set of unique spaces (e.g. base or PR branch).

Also, I'm a bit unsure about what `default_space` means in this context. Assuming that it would e.g. be the "default branch" of a Git repository, I'm not sure there are many other use cases where a `default_space` can be determined?

vito: @itsdalmo I think that's cleaner, but the one disadvantage is that it won't be able to return spaces that have no versions. Though perhaps that's not very useful anyway.

The notion of `default_space` would be used for two things:

1. In the `git` resource (probably any VCS resource), this would return the default branch.
1. For semver'd resources (e.g. Docker images, S3 artifacts, ...) this would return the space corresponding to the latest version. This is to match today's behavior where by default you may want to just "always roll forward".

mhuangpivotal:
We implemented a v2 interface proof-of-concept for the s3 resource. The code can be found [here](https://gist.github.com/mhuangpivotal/c647417fb7614e33f7930a99324744e7).

01-resources-v2/proposal.md
```diff
@@ -137,40 +202,99 @@ type PutRequest struct {
...
+jobs:
+- name: run-pr
+  plan:
+  - get: concourse-pr  # pr: 123, ref: deadbeef
+    trigger: true
+  - get: concourse     # ref: abcdef
+  - put: concourse-pr
+    get: merged-pr
+    params:
+      merge_base: concourse
```
itsdalmo: Ref your comment in https://github.com/telia-oss/github-pr-resource/issues/5#issuecomment-440416235, I figured I'd leave my questions here! 

Given the documentation in the RFC, presumably we must have a `default_space` in the source configuration for our resources, or specify `spaces: []` for each `get`:

```yml
- name: run-pr
  plan:
  - get: concourse-pr
    spaces: [ master ] # <- Must be configured here or in the source right?
    trigger: true
  - get: concourse     
    spaces: [ master ] # <- Same
  - put: concourse-pr
    get: merged-pr
```

And based on the above/If i've understood your suggestion correctly, it seems that with Resources V2 the `github-pr-resource` should emit versions like this:


```json
{
  "action": "discovered",
  "space": "master",
  "version": { 
    "pr": 123,
    "ref": "deadbeef"
  },
  "metadata": [{...}]
}
```

This necessarily means that our job `name: run-pr` can only ever test PRs against a single branch at a time. So if someone is running e.g. GitFlow or something similar with multiple branches, they would have to add a job for each base (branch, e.g. `develop` and `master`)? 

This assumes that we **can't** do `spaces: [develop, master]` on `get`, but even if we could, we'd also need complex logic in Concourse to figure out that a PR against `develop` should use the latest version from the `develop` space of `get: concourse` when creating `merged-pr` - which I presume is not there?

itsdalmo: PS: The last paragraph before the "example"-section states that:
> the user must specify it explicitly in the pipeline, either by configuring one on the resource (default_space: foo) or on every get step using the resource (spaces: [foo]).

I believe `spaces: [foo]` should be `space: foo` according to the documentation for `get` further down? I'll update my comment above to reflect this.

vito: @itsdalmo `get` steps specify arrays of spaces (and/or patterns like `feature/*` for all `feature/`-prefixed branches). You could specify `[develop, master]` there but I'm not sure that'd be very useful. Sorry, I'm having kind of a hard time parsing your question - is the concern that the base branch for a PR varies from PR to PR and you'd need some way to coordinate that with this approach?

001-resources-v2/proposal.md
```diff
@@ -0,0 +1,626 @@
...
+}
+```
+
+The command will be invoked with a completely empty working directory. The
+command should populate this directory with the requested bits. The `git`
+resource, for example, would clone directly into the working directory.
+
+If the requested version is unavailable, the command should exit nonzero.
+
+No response is expected.
```
vito: After discussing with @clarafu we think it'd actually be better if `get`s still returned a response, and have that response include metadata.

This makes `get`, `check`, and `put` all have a response, and lets the resource author decide where metadata should be discovered. Now the difference between v1 and v2 would be that with v2 `check` can *also* return metadata - an additive change, rather than requiring all metadata come from `check`.

This way if there are situations where it's too expensive to collect all the metadata at once, it can be deferred until later `get` and `put` calls. Any metadata returned by `get` and `put` would just update the version's metadata.

This also makes the implementation in the ATC a bit cleaner because we don't have to worry about a scenario where e.g. a `get` is running with a version that has never had a `check` run and so there's no metadata to show to the user in the UI. This normally won't happen but is theoretically possible via the builds API, where you can submit a build that fetches an arbitrary version. This also makes v1 and v2 more similar to each other, minimizing conditional code paths in the implementation.

001-resources-v2/proposal.md
```diff
@@ -0,0 +1,626 @@
...
+  interfaces, their protocol versions, and any other interface-specific
+  meta-configuration (for example, which commands to execute for the
+  interface's hooks).
+
+* The first supported interface will be called `artifacts`, and its version
+  will start at `2.0` as it's really the next iteration of the existing
+  "resources" concept, but with a more specific name.
+
+* There are no more hardcoded paths (`/opt/resource/X`) - instead there's the
+  single `info` entrypoint, which is run in the container's working directory.
```
marco-m: `info` semantically implies read-only. If it is going to replace also the `out`/`put`, maybe a more adequate name could be `command` or similar.

001-resources-v2/proposal.md
```diff
@@ -0,0 +1,626 @@
...
+
+* Change `get` script to no longer return a version, since it's always given
+  one now. As a result, `get` no longer has a response; it just succeeds or
+  fails.
+
+* Change `get` and `put` to run with the bits as their working directory,
+  rather than taking the path as an argument. This was something people would
+  trip up on when implementing a resource.
+
+* Change `check` and `put` to write its JSON response to a specified file,
```
marco-m: its -> their
response -> responses

001-resources-v2/proposal.md
```diff
@@ -0,0 +1,626 @@
...
+  rather than `stdout`, so that we don't have to be attached to process its
```
marco-m: > so that we don't have to be attached

this is unclear. Who is attached to what ?

jchesterpivotal:
A question I've thought of whether resources could add a `provision` capability as well. There are some resources where the remote service is ephemeral or at least unimportant to humans. Examples would be registering a webhook with Github, creating randomly-named blobstore directories, or creating and binding databases. I think OSBAPI _might_ be able to serve in this role.

Arguing against: any `provision` could instead be modeled as a `put` to a provisioning resource. Existing examples including the `bosh-deployment` resource, which essentially provisions systems intended for further usage.

The inspiration for this thought bubble is watching over the shoulder of Knative Eventing. In Concourse terms, I think Knative Eventing somewhat mixes together the concerns of provisioning an outside service with configuration of `source`/`params`.

001-resources-v2/proposal.md
vito: Confusing wording - it's read-only, I meant `entrypoint` from a discovery standpoint (invoke `info` to learn the later commands to run), not an execution standpoint. Will amend.

vito:
I've been doing a lot of thinking about this lately, and now I'm worried that baking a high-level concept like 'spaces' into the resource interface is the wrong direction. I think I've got a better idea that simplifies things down to a single resource interface, a *lot* like what we have today - and a plan to use this single interface to implement spaces, triggers (e.g. the `time` resource), notifications (e.g. Slack, GitHub status checks), and versioned artifacts (today's primary use of the interface).

There's a lot to go over here - I'm going to try and provide as much context as possible behind my thought process. I expect at least some of this to be confusing or unclear. Sorry in advance for the word vomit. 😅 Feel free to ask questions to clarify - I plan to open a new RFC as it's a significant enough departure from this proposal, so we might as well just open the discussion up here until then.


## Triggers and the stampeding herd problem

With concourse/concourse#2386 we formalized how resource configs and their versions are stored abstractly, but this introduced a problem: now the `time` resource would cause a stampeding herd, as everyone who said `interval: 10m` would get the same history, so all their jobs would fire at once across the cluster. We fixed this by the introducing 'resource scopes' which allow resources to have their own version history exclusive to their pipeline resource definition.

Resource scopes proved challenging to implement and introduce a level of overhead that is only exacerbated by the addition of resource *spaces* to the model, as they both try to achieve similar things. This led us to try to rethink the `time` resource as an example of *inbound* notifications, or "triggers". The hope was by switching `time` to a different concept and interface we would get rid of the problem of 'shared version history' as we would just stop treating it like an artifact. Context: https://github.com/concourse/concourse/issues/3585 and https://github.com/concourse/concourse/issues/3595

But then I thought maybe we're over-complicating things by introducing a whole new interface. If you look at the `time` resource's implementation, what it's doing pretty much makes sense in isolation. It's just that we don't want to centralize the history - in fact we don't even really want to *store* the history. Maybe we just need a different way to use resources in a pipeline in this way?

Coincidentally, @chenbh opened https://github.com/concourse/concourse/issues/3572 around the same time we were thinking about this. What's interesting is here they still want to use a resource, but only as a trigger, and they don't want to fetch the bits. Sounds familiar! 🤔 

That led me to think maybe the *interface* is fine, and we just need different ways to use resources in a pipeline. Maybe everything *is* just "resources" in the sense that they're all the inputs to your pipeline and still represent all external state, but maybe we just need to allow your pipeline to have different relationships with this state? Maybe everything just isn't *versioned artifacts*?

It's worth noting briefly that the stampeding herd problem still exists within a pipeline *today*, not just with globally shared history. Having all the jobs that point to a time trigger all fire at once doesn't seem great.

So what if you could associate a resource to a job, and have its `check`s always be relative to the last-used trigger "version" that the job used? That way each job naturally has its own interval, and we're still just using the `time` resource as it is. The stampeding herd problem is then solved at all levels, without the need for a new interface.


## Spatial resources

If we go with v2 RFC as-is, all resources implicitly have to care about spaces. What does this 'trigger-only' workflow even mean for spaces? ...Should spaces just be optional?

What if spaces wasn't part of the interface at all? Just like we were able to implement triggers in terms of the resource interface, can we implement spaces in terms of it, too?

The meaning of `check`, `get`, and `put` map pretty well to the spaces themselves:

* `check`: returns a 'version' for each space. We would just never give it a cursor version, and do a set diff against the result. Users can configure whatever filters they want in `config` to control which spaces are returned.
* `put`: supports creation/deletion of spaces (i.e. git branches).
* `get`: returns whatever data is useful for a given space.

What's nice is this makes the detecting of spaces a lot clearer, and supports explicit creation/deletion of the spaces themselves. The current proposal puts all of this into `check` which checks across all spaces and has to emit a `reset` event when one disappears. With this interface, the space is just no longer returned.

So, how would we use this to support spaces over an artifact resource like `git`?

We take the "versions" returned by the space `check` and toss them in the config of the artifact resource, like so:

```
space /check:
    {"uri":"..."}
 -> {"branch":"foo"}
    {"branch":"bar"}

artifact /check:
    {"uri":"..."} + {"branch":"foo"} = {"uri":"...","branch":"foo"}
 -> {"ref":"abcdef"}
    {"ref":"deadbeef"}
```

In this case, the `git` resource already understands `branch:` as part of its `source:` config, so it doesn't even have to change! I think this is common to lots of existing resources, actually - the things they would have been changed to "space" over are actually things they already support being pointed to in their `source:` config.

The interesting thing here is it uses the 'version' as a way to compose resources together. It actually feels pretty nice! Versions are already a public contract and are even shown to users directly in the UI.

What's *really* interesting about this is that by decoupling the space resource from the artifact resource, you could space over arbitrary things. At the end of the day you're just dynamically providing fields to `source:`. If you wanted to have a space for testing different kinds of arbitrary configuration, not just what the resource author predicted, you can do it by writing your own space resource.

This also removes the whole idea of a "default space". Now that resources don't have to implement spaces, it's not necessary to even think about as a resource type author. Either your `source:` is statically configured or dynamically modified via a space resource - as an author you don't know or care.

Another thing that's great about this is that, internally, everything is just back to resource configs and their versions. No more scopes, spaces, etc. - just resource configs and their associated versions. In this relationship, one resource config just results in another resource config being created.

## Notifications

An early iteration of this RFC actually introduced notifications as a new part of the v2 interface, and they could be tied to artifact resources. This was so that things like GitHub status checks could be sent back with an associated version, so the notification resource knows what commit to apply the status check to.

So, what if we don't add a new interface for notifications, and we just use the 'version' as a communication mechanism between resources, just like we did for spaces? What if you could point a notification resource at an artifact resource, and have Concourse automatically run the notification resource when the artifact resource is used in a build?

This would work by taking the version of the artifact resource and tossing it in the `config` when invoking the `put` action of the notification resource, like so:

```yaml
{
  "config": {
    # arbitrary user-specified config
    "repo": "concourse/concourse",

    # taken from the monitored resource
    "ref": "abcdef"
  },

  # provided by concourse
  "build": {"id": 123, "status": "failed", "job_name": "some-job", "url": "https://..."}
}
```


## Overall proposal

I think we should go back to resources being a super simple and general `check`/`get`/`put` interface, and cut spaces out of the proposed interface. I think this RFC has run its course and I can open a new RFC with the simpler proposal. This new RFC would be a more incremental change on the existing interface, and it would still include things like version deletion and collecting all versions back in time. With spaces out of the picture we can finally make these changes without putting all our eggs in one basket.

The new proposal would introduce the general interface, and then describe how it would be interpreted to support 'versioned artifacts', as the interface is used for today, and also touch on the ideas around composability I covered above, ultimately so that the common interface isn't too tied to versioned artifacts. (For example, we may want to use a name other than 'version'.)

The challenging part of this approach is how we bubble these concepts up to users. The pipeline has to feel intuitive and not too repetitive. It shouldn't take a degree in YAML architecture to configure GitHub status checks. But the more I think about this, the more apparent the benefits become in terms of the mental model and internal structure. I think having a consistent model is paramount for Concourse's success (it's one of the founding principles).

Here are some key take-aways summarizing the general direction I think we should go:

* Resources are general interfaces for interacting with external state.
* Concourse pipelines define the relationship your pipeline has with this state.
  * versioning artifacts
  * emitting notifications
  * triggering jobs
  * fanning out across branches dynamically
  * configuring build matrixes
* Resources receive a single `config` and can emit config fragments (née versions).
* Resources can compose with each other by passing config fragments to each other.
  * Config fragments can feed into `/check`.
    * Space `{uri}` config -> `{branch}`...
    * Artifact `{uri}` config + space `{branch}` fragment -> `{ref}`...
    * Artifact `{uri,branch}` config + artifact `{ref}` fragment (of last seen version) -> `{ref}`...
    * Or: artifact `{uri}` config + space `{branch}` fragment + artifact `{ref}` fragment (of last seen version) -> `{ref}`...
  * Config fragments can feed into `/get`.
    * Artifact `{uri,branch}` config + artifact `{ref}` fragment = `git clone {uri}; git checkout {ref}`
  * Config fragments can feed into `/put`.
    * Notification `{repo}` config + artifact `{ref}` fragment = update GitHub status
* Add pipeline functionality for trigger-only resources as another RFC.
  * These are resources whose `checks` run relative to their associated job's last use of the notification.
* Add pipeline functionality for notifications as another RFC.
  * These are resources that are sidecar-ed onto another resource and follow its path through the pipeline.
* Add pipeline functionality for spaces as another RFC.
  * These are resources that are referenced in tandem with an artifact resource to run against each of the config fragments returned by the space. Something like `{get: foo, spaces: bar}`.
  * Spaces are a pipeline-only add-on feature that users opt in to by composing resources explicitly.
  * Now we don't have to worry about backwards-compatibility, because now the jump from v1 to v2 is easy, and spaces don't fundamentally change the resource model.

Critically, those last 3 RFCs can be done with both the v1 and v2 interfaces. It's not so coupled anymore. If you subtract spaces from resources v2, a lot of it is boring protocol changes, plus things like "`put` can create/delete many versions". That all is pretty independent of the other RFCs. It's a bit clearer with v2, of course, since v2 has just "config", but for v1 we can say that 'versions' (config fragments) are shoved into 'source' (config).

vito:
Proposing we close this in favor of https://github.com/concourse/rfcs/pull/24, which is still a work in progress but at least the resource interface part is pretty close. :smile: I'll open separate RFCs for spaces/triggering/notifications after the resource interface portion of #24 is done. (Right now they're all in there together.)

jchesterpivotal:
Will the lengthy comment above somehow come across to the new proposal? It carries a lot of important context and food for thought.

vito:
@jchesterpivotal The new proposals will encompass everything covered there, yes, though at the moment in a bit of a dry sense - the RFCs just outline the 'what' whereas my comment earlier was kind of a brain dump/train of thought that led to it. Is there any particular part you feel I should make sure to incorporate? Maybe they could just reference my comment for context-building? :thinking:

jchesterpivotal:
Perhaps as a discussion document?

I'm concerned that lots of folks will overlook a detailed discussion on a closed PR.

vito:
@jchesterpivotal I've added a 'Previous Discussions' section to the new RFC: https://github.com/concourse/rfcs/pull/24/commits/0a805ec8425344ff1e5202a707b9c42d706c4388

vito:
Closing in favor of #24. I was gonna wait on the post announcing it per the [RFC resolution process](https://github.com/concourse/rfcs#resolution) but then realized folks can just close *their own* proposals without having to go through the resolution process. :sweat_smile: I'm writing a post for #24 anyway though and will mention the closing of this one.

PR #2, opened 2018-04-17
RFC: Kubernetes Runtime

topherbullock:
[Proposal](https://github.com/topherbullock/rfcs/blob/k8s-runtime/02-k8s-runtime/proposal.md)

Status: draft

Still more details and thoughts to collect; specifically around how the new GC changes impact how a K8s runtime should clean up resources. I'm sure more questions will come up around using K8s effectively.

Please submit comments and feedback on individual lines of the proposal markdown, so that we can track open conversations.

### DIFF
--- a/02-k8s-runtime/proposal.md
+++ b/02-k8s-runtime/proposal.md
@@ -1,6 +1,6 @@
 # Summary
 This RFC outlines the concerns and considerations involved in introducing a
-Kubernetes runtime, which schedules Kubernetes Resources in place of Garden
+Kubernetes (K8s) runtime, which schedules K8s Workloads in place of Garden
 Containers and (potentially) Baggageclaim Volumes.
 
 # Proposal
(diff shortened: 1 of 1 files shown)


02-k8s-runtime/proposal.md
```diff
@@ -0,0 +1,193 @@
...
+## Multiple Worker Orchestrator "Types"
+
+Operators of a Concourse cluster may need to be aware of the underlying
+orchestrator their cluster is using, or the "type" of registered workers.
+
+## Worker Tagging
+
+Tagging of Concourse Jobs for specific workers might need to change to
+accommodate K8s Pod `nodeSelector`s which allow users to select specific K8s
+Nodes schedule the necessary Workloads on. Affinity and anti-affinity also
```
edude03: IMO, neither affinity or anti affinity should be exposed to the user.

ornous: Hey edude03. Why do you believe that?

ornous: Hey edude03. Why do you believe that?

02-k8s-runtime/proposal.md
```diff
@@ -0,0 +1,193 @@
...
+depending on users' pipelines. The contract with Concourse users is that
+inputs will be mounted as a Copy(OnWrite) of the Resource Cache Volume, and
+modifications to the input within a task will not mutate the contents of the
+cache.
+
+Is there a more "K8s way" of managing volumes between Pod executions, which
+provides and interface the ATC can leverage to manage volumes similar to how
+Baggageclaim Volumes are managed. Could a replica set place a worker on each K8s
+Node and use `HostPath` Volumes or create Persistent Volumes to store Resource
+Caches, Task Caches, and Image Resource Caches, etc.?
```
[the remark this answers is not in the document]
vito: @edude03 I don't think I quite understand persistent volume claims enough yet to know what you mean by the state desync, but I can theorize. The gist of things is that the set of volumes a worker has is associated *by name* to the worker's registration in the ATC's database. If a worker comes back with a new name but is somehow pointing to a disk with a bunch of volumes already present, the ATC won't know they exist. Similarly, if a worker comes back under the same name but all of its volumes are gone, the ATC will get a bit confused and still try to use them.

As far as BaggageClaim is concerned, its API is just a direct reflection of the filesystem (a `GET` request to list volumes corresponds to a directory listing for example). It has no idea when its volumes are mounted/unmounted to/from containers, it's just preparing the paths on the host that we then bind-mount. Its sole purpose is to manage the volumes through direct API calls from the orchestrator. There's no state in its head and no automatic garbage collection (beyond volume TTLs, which aren't used by Concourse anymore but are still implemented in the API) or outbound interaction with the ATC to "re-sync".

As for how BaggageClaim fits in to the K8s model, I wonder if it would be a good fit as a [CSI](https://kubernetes.io/blog/2018/01/introducing-container-storage-interface/)? Maybe this way we could abstract over whether the volume has to be transferred over the network or a local copy-on-write can be constructed. (Disclaimer: I've only briefly looked at this...)

02-k8s-runtime/proposal.md
edude03: I think that it makes the mental model of workload scheduling more confusing. At the very least it shouldn't be part of an MVP

02-k8s-runtime/proposal.md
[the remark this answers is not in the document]
edude03: @vito - so I'm going from memory so the situation might have changed however using the Kubernetes helm chart problem the first problem is that the workers crash fairly frequently. 

This leads to / exasberates the second problem which is that I get an error about not being able to find the volume causing the worker to get stuck - it's running but none of the jobs assigned to it actually run. 

I've never understood this problem because the helm chart is implemented as a `StatefulSet` which means that the worker has a stable name (it will always be called `concourse-worker-$n`) and it uses a `PersistentVolumeClaim` so it will restart with the same "Hard drive" (I'm avoiding calling it a volume as that term is overloaded in this context) as the previous run. 

With this setup - it should have the same guarrentees as if run on a physical machine (host scheduling not with standing of course) and yet workers break fairly often. 

As for the statefulness of ATC/BaggageClaim - it seems to me that one solution to this would be to make volume requests an upsert operation IE create the volume if it doesn't exist. OR to remove the cleverness of trying to schedule jobs on machines were the volume already exists. Of course figuring out why volumes can't be found on Kubernetes is probably an even better solution.

As for a CSI - It's either a great idea or not the right thing at all though I'm not sure I understand baggageclaim enough to say. 

a CSI can run as a sidecar in a pod, listening to the kubernetes API and respoding to requests for storage (Add/Update/Delete operations).

 This sounds like what baggageclaim is for although I imagine a Pod would need a `PersistentVolumeClaim` mounted so that baggageclaim has somewhere to put its own volumes?

 In the K8S model, would baggageclaim map one "Volume" to one "hard drive" or would it get one "hard drive" and create a `btrfs` image and put all the volumes in there?

pivotal-nader-ziada:
One approach I investigated for Volume management is the Local Persistent Volumes feature. Each worker node would have a local persistent volume, and a persistent volume claim, that be be then mounted in the Pod executing a given task. PersistentVolume nodeAffinity enables the Kubernetes scheduler to schedule Pods using local volumes to the correct node.
On this local persistent volume, folders represent each required volume and the subPath is mounted in the Pod that needs access to the volume. Here is a note about the subpath details and fixes to a recent vulnerability.
https://kubernetes.io/blog/2018/04/04/fixing-subpath-volume-vulnerability/ 

An example for creating local persistent volume in the following yaml files
https://gist.github.com/pivotal-nader-ziada/15d430bb16397c672e337f2e9275164c 

This solution requires significant changes to how concourse is currently managing volumes, and at the same time is not using a kubernetes feature as is.

02-k8s-runtime/proposal.md
topherbullock: @edude03 @ornous 
tags already allow Concourse users to define some aspects of how steps are scheduled across the set of workers, but those are all exposed by workers themselves

It would be confusing for users to need to reason about what nodes their resources are scheduled on, but maybe the CRD for workers could tie the thread from worker tags -> node.

hairyhenderson:
Just wanted to drop this here: https://blog.drone.io/drone-goes-kubernetes-native/ - discusses how Drone (a similar CI/CD tool) has leveraged Kubernetes. Some of the implementation details are potentially relevant (or not!)

andrewedstrom:
Glad to hear it! This is a really interesting idea. Has anyone tried a spike of this? Or is it such a huge undertaking that such a spike doesn't even seem feasible.

pivotal-jwinters:
@topherbullock As part of this investigation it might be interesting to consider what a garden backend  for kubernetes might look like, and how it would fit into our lifecycle. Also baggage claim would need ...a new driver? ...something higher level? We would likely have to rework some stuff but it could be an interesting approach.

ralekseenkov:
I personally think that native integration with Kubernetes is a lot more work, but it will open way more doors for Concourse in the open-source community.

going k8s garden backend + k8s baggage claim driver route might be technically feasible. but people may still not treat it as "Kubernetes native". it might add more points or failure. and likely the experience of people who run Concourse on Kubernetes will still be suboptimal.

in other words, it would be great if we can look at it as "**how can we make Concourse # 1 go-to option for people who run CI/CD on Kubernetes**" as opposed to "just adding Kubernetes support to Concourse".

just my 2c... hopefully this makes sense :)

pivotal-jwinters:
I guess it all depends on what the "native kubernetes abstraction" looks like. It could be that the garden backend interface is too restrictive and we lose out on some useful native k8s features. But if at the end of the day all we're doing is creating `volumes` and `tasks` and handing it off to the k8s runtime then there may not be much of a difference in the two approaches. 

But yeah I totally agree that we want to provide a better user experience than "just a shim". Plus it would be really cool if whatever we come up with could be extensible enough so that we can support a concourse docker runtime.

topherbullock:
> Glad to hear it! This is a really interesting idea. Has anyone tried a spike of this? Or is it such a huge undertaking that such a spike doesn't even seem feasible.

@andrewedstrom Yeeeaaaah... implementing this is a giant leap rather than a few small steps.  As a conversation piece this RFC has served its purpose for now, and a v2 would cover some more specific, actionable steps to leverage **_[something that isn't Garden]_** for Containers and Volumes.

>  As part of this investigation it might be interesting to consider what a garden backend for kubernetes might look like, and how it would fit into our lifecycle. 
- @pivotal-jwinters 
> I personally think that native integration with Kubernetes is a lot more work, but it will open way more doors for Concourse in the open-source community.
- @ralekseenkov 

I agree that we will need to consider the K8s runtime at a higher level than the granularity we have with Garden for now. We're ( _likely_ ) dealing with a higher order of abstractions in K8s (Pods/Deployments/Jobs , rather than containers directly ) , and we don't have a fine-grained way to control Volumes and their lifecycle separate from the Pod our container(s) are a part of. 

Concourse has evolved a lot on this path thanks to recent discussions about the [design of the "Runtime"](https://github.com/concourse/concourse/issues/2926) and the interface between Runtime and Core,  [new scheduling strategies](https://github.com/concourse/concourse/issues/3301), and the investigation of [ephemeral check containers](https://github.com/concourse/concourse/issues/3079). I feel like we're slowly moving toward a path where the "runtime" has a higher-order interface which is leveraged by the business logic in Core+API, which warrants a separate RFC for "What does a unified interface for a swappable runtime look like?" separate from K8s discussions.

Knative has spawned [a separate initiative for CI/CD](https://github.com/knative/build-pipeline), which means the conversations around the K8s runtime specifically have also evolved beyond this RFC.

topherbullock:
Closing and creating 2 new discussions from this RFC : 
- K8s / Knative / Tekton
- Swap-able Runtimes

PR #3, opened 2018-05-04
First cut at refactoring proposal for domain model persistence

jmelchio:
Starting spike to implement on one model object.


vito:
I don't think this is going to be merged at this point (it was made quite early on in the RFC repo's life before we figured out what we're doing with it), but a lot of the advice is sound. We've since introduced a [Concourse Go Style Guide](https://github.com/concourse/concourse/wiki/Concourse-Go-Style-Guide) wiki which touches on a lot of this (largely by deferring to existing style guides).

PR #6, opened 2018-07-18
RFC: RBAC

pivotal-jwinters:
[Proposal](https://github.com/pivotal-jwinters/rfcs/blob/proposal/rbac/003-rbac/proposal.md)

Status: Draft

This doesn't seem like that big of a change, and I think most proposed changes are fairly straight forward. I'm not a huge fan of my proposed solution for `fly set-team`, but with our static flags (i.e. `--github-user`) being as they are, I can't really see how else this would work.

Or you can view the [original issue](https://github.com/concourse/concourse/issues/2389)


xtreme-sameer-vohra:
Copying from #2389

As an alternative to require 2 commands to set some `viewers` & `members`, could we also consider alternatives that might allow configuring a team with a single command. 
This would remove the point mentioned above
```
The main drawback here is that we would have to merge auth configuration because viewers and members would be configured in two separate steps. The current behaviour overwrites the entire auth config each time, so in keeping with the current behaviour I guess we would overwrite the entire config for each role separately.
```

For example set-team could allow you to specify a config file with the following content
```
{
	"viewer": {
		"groups": [],
		"users": []
	},
	"member": {
		"groups": ["github:myorg:myteam"],
		"users": ["github:pivotal-jwinters"]
	}
}
```

xtreme-sameer-vohra:
For the team viewer role, should they be able to access the team resource itself including the auth config ?
That might be an operation that we might restrict to team members only.

03-rbac/proposal.md
```diff
@@ -0,0 +1,110 @@
...
+fly -t mytarget set-team -n myotherteam --github-user pivotal-jwinters --github-team myorg:myteam
+```
+
+We could allow roles to be set as follows:
+```bash
+fly -t mytarget set-team -n myteam --role viewer --allow-all-users
+fly -t mytarget set-team -n myteam --role member --github-user pivotal-jwinters --github-team myorg:myteam
+```
+
+The main drawback here is that we would have to merge `auth` configuration because `viewers` and `members` would be configured in two separate steps.  The current behaviour overwrites the entire auth config each time, so in keeping with the current behaviour I guess we would overwrite the entire config for each role separately.
```
vito: I think this makes sense, but at this point maybe we should explore having team config files instead, and having it set everything at once, including all the roles. Kind of like Vault's policy config files: https://www.vaultproject.io/docs/concepts/policies.html (in spirit)

We could then allow configuring multiple teams on ATC start by pointing it at the config files. This would even make the `main` team no longer "special" - it'd just be among the set of teams configured at start, with `admin` set to `true`.

pivotal-jwinters: @xtreme-sameer-vohra I'm wondering if your exported team list should match the config format that we choose?

xtreme-sameer-vohra: @pivotal-jwinters Yep, it makes sense for `fly teams` to be able to export team contents in the same format such that it could be used for `set-teams` as well/

xtreme-sameer-vohra:
Would we consider `seeded` teams ( created via the manifest ) static ? 
If we allow users to also update a particular team that was configured in the manifest by using `fly set-team` or `fly set-teams` then it will be hard to infer the desired behaviour.

vito:
@xtreme-sameer-vohra yep, i think so: https://github.com/concourse/concourse/issues/1193#issuecomment-393195930

jama-pivotal:
This is the permission matrix I've written up to capture both fly and web view actions for each potential member. It's not comprehensive but should cover most use cases.

https://docs.google.com/spreadsheets/d/1np3hyJy3mVRfB2gcgKykz3QTQg5qEj28QgK523SEmao/edit#gid=0

jchesterpivotal:
Dumb question: will this overlap with or cause confusion vs Kubernetes RBAC?

03-rbac/proposal.md
```diff
@@ -0,0 +1,159 @@
...
+```bash
+fly -t mytarget set-teams -c /tmp/team-config
+```
+
+Where `/tmp/team-config` might look something like:
+
+```yaml
+teams: 
+  main: 
+    admin: true
```
ebabani: Should this be configurable for the `main` team? What's the recovery if someone makes this false?

03-rbac/proposal.md
```diff
@@ -0,0 +1,159 @@
...
+
+- What roles do we need?
+- How would fly set-team work?
+- How do roles get persisted in the database?
+- How are roles handled by the API?
+
+
+# Proposal
+
+## What roles do we need?
```
william-tran: Here are our needs, at a minimum:
- dashboard - we have a custom dashboard. It needs GET access to the following and no more, so that the credentials that the dashboard uses can't be used to get build logs that could leak other credentials.
  - `/api/v1/teams/{team}/pipelines`
  - `/api/v1/teams/{team}/pipelines/{pipeline}/jobs`
  - `/api/v1/teams/{team}/pipelines/{pipeline}/resources` (to display check failures)
  - `/api/v1/teams/{team}/pipelines`
  - `/api/v1/teams/{team}/builds`
- user - everything in their team, except `set-team`, `hijack`, `execute`, `set-pipeline`. We want to ensure every action that concourse does is auditable. We have a pipeline that gets pipeline configs from git, and does the set-pipeline-ing, and we need to ensure that is the only way that pipelines can be set
- admin - everything

Because these roles and permissions are quite a bit different, and I'm sure others will have their own idea of what they should be, it would be great if this was configurable.

william-tran:
Here here, I'm hoping we get the ability to define our own roles, this would save us a lot of work in writing our own proxy with auth etc.

pivotal-jwinters:
@Rukenshia @william-tran our first pass at RBAC is going to have static roles, but we have started thinking about what dynamic roles might look like.

The thought (similar to what @Rukenshia mentioned) is that down the road you'll be able to map specific actions to whatever you want, but our initial action/role mapping will probably look something like the following: 

```
	atc.DestroyTeam:                   "admin",
	atc.RenameTeam:                    "admin",
	atc.SetTeam:                       "admin",
	atc.AbortBuild:                    "member",
	atc.CheckResource:                 "member",
	atc.CheckResourceType:             "member",
	atc.CheckResourceWebHook:          "member",
	atc.ClearTaskCache:                "member",
	atc.CreateBuild:                   "member",
	atc.CreateJobBuild:                "member",
	atc.CreatePipelineBuild:           "member",
	atc.DeletePipeline:                "member",
	atc.DeleteWorker:                  "member",
	atc.DisableResourceVersion:        "member",
	atc.EnableResourceVersion:         "member",
	atc.ExposePipeline:                "member",
	atc.HeartbeatWorker:               "member",
	atc.HidePipeline:                  "member",
	atc.HijackContainer:               "member",
	atc.LandWorker:                    "member",
	atc.OrderPipelines:                "member",
	atc.PauseJob:                      "member",
	atc.PausePipeline:                 "member",
	atc.PauseResource:                 "member",
	atc.PruneWorker:                   "member",
	atc.ReadOutputFromBuildPlan:       "member",
	atc.RegisterWorker:                "member",
	atc.RenamePipeline:                "member",
	atc.ReportWorkerContainers:        "member",
	atc.ReportWorkerVolumes:           "member",
	atc.RetireWorker:                  "member",
	atc.SaveConfig:                    "member",
	atc.SendInputToBuildPlan:          "member",
	atc.SetLogLevel:                   "member",
	atc.UnpauseJob:                    "member",
	atc.UnpausePipeline:               "member",
	atc.UnpauseResource:               "member",
	atc.BuildEvents:                   "viewer",
	atc.BuildResources:                "viewer",
	atc.DownloadCLI:                   "viewer",
	atc.GetBuild:                      "viewer",
	atc.GetBuildPlan:                  "viewer",
	atc.GetBuildPreparation:           "viewer",
	atc.GetConfig:                     "viewer",
	atc.GetContainer:                  "viewer",
	atc.GetInfo:                       "viewer",
	atc.GetInfoCreds:                  "viewer",
	atc.GetJob:                        "viewer",
	atc.GetJobBuild:                   "viewer",
	atc.GetLogLevel:                   "viewer",
	atc.GetPipeline:                   "viewer",
	atc.GetResource:                   "viewer",
	atc.GetResourceCausality:          "viewer",
	atc.GetResourceVersion:            "viewer",
	atc.GetVersionsDB:                 "viewer",
	atc.JobBadge:                      "viewer",
	atc.ListAllJobs:                   "viewer",
	atc.ListAllPipelines:              "viewer",
	atc.ListAllResources:              "viewer",
	atc.ListBuilds:                    "viewer",
	atc.ListBuildsWithVersionAsInput:  "viewer",
	atc.ListBuildsWithVersionAsOutput: "viewer",
	atc.ListContainers:                "viewer",
	atc.ListDestroyingContainers:      "viewer",
	atc.ListDestroyingVolumes:         "viewer",
	atc.ListJobBuilds:                 "viewer",
	atc.ListJobInputs:                 "viewer",
	atc.ListJobs:                      "viewer",
	atc.ListPipelineBuilds:            "viewer",
	atc.ListPipelines:                 "viewer",
	atc.ListResourceTypes:             "viewer",
	atc.ListResourceVersions:          "viewer",
	atc.ListResources:                 "viewer",
	atc.ListTeamBuilds:                "viewer",
	atc.ListTeams:                     "viewer",
	atc.ListVolumes:                   "viewer",
	atc.ListWorkers:                   "viewer",
	atc.MainJobBadge:                  "viewer",
	atc.PipelineBadge:                 "viewer",
```

pn-santos:
We just came across a use case that I'm unsure if it could be solved with the planned RBAC implementation (or by some other means). So we have a job that will require credentials that we don't want exposed to the users that will be making use of the job (use means triggering and seeing its log output). Basically we want to prevent/limit the users who can intercept the container that will run the job.

If I'm reading the proposal correctly, in order to grant "triggering" rights a user also have the ability to  intercept the container.

Does preventing interception but still allow a user to trigger the job make sense? would it be feasible/possible with the RBAC system? (or in some way possible already?)

03-rbac/proposal.md
```diff
@@ -0,0 +1,186 @@
...
+  cf: 
+    users: ["myusername"]
+    spaces: ["myorg:myspace"]
+- name: viewer
+  allow_all_users: true
+```
+
+or this:
+
+### ✅ Option 2 - roles map
```
aegershman: Just want to say that I like this quite a bit. Git-backed configuration is incredibly powerful, and it allows self-service for Concourse access: it allows teams within the company to create PRs against the config repo, and then CI can make the changes once things get merged into master.

elgohr:
I'm missing technical users in the discussion. We often do pipe of pipes. In this way there would be the need for a "technical member". I don't know whether this would make a difference to the normal member.
I'm thinking of something like Github Deploy Keys.

vito:
Hi all, we think this is good to go at this point and I'm going to merge it in (as our first merged RFC!)

This is jumping the gun on the normal RFC process a bit because this RFC predates the solidification of the process itself. Normally we would go through a more extended review before accepting it, however the implementation for this has already been done in `master` for months, so there's not much point.

There are a couple outstanding questions left but I think those can come up as later issues to improve on top of the RBAC work established by this RFC and initial implementation. Thanks to everyone that took part in this!

PR #8, opened 2018-08-20
RFC: Merge resources and resource_types

clarafu:
[Proposal](https://github.com/clarafu/rfcs/blob/master/05-recursive-resources/proposal.md)

Status: Draft

The issue that introduced this idea: [github-issue-link](https://github.com/concourse/concourse/issues/2386)

Looking to collect some input from the community about this idea, as we are still at the beginning of drafting it out and deciding if this is something we want to do. Please submit comments and feedback on individual lines of the proposal markdown, so that we can track open conversations.

Thanks! :)

### DIFF
--- a/05-recursive-resources/proposal.md
+++ b/05-recursive-resources/proposal.md
@@ -1,0 +1,21 @@
+# Summary
+
+Merging resource types into resources:
+
+* Easily introduces visibility into resource types (storing history of versions, pinning a version, etc).
+
+* Removes the need for duplicate code (example, supporting fly check-resource-type and fly check-resource).
+
+* Allows for easier process for creating a new resource type (can create a pipeline that runs tests on new versions of a resource type)
+
+
+
+# Proposal
+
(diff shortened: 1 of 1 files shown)


05-recursive-resources/proposal.md
```diff
@@ -0,0 +1,21 @@
...
+* Removes the need for duplicate code (example, supporting fly check-resource-type and fly check-resource).
+
+* Allows for easier process for creating a new resource type (can create a pipeline that runs tests on new versions of a resource type)
+
+
+
+# Proposal
+
+Currently, Concourse treats resources and resource types quite similarly. This proposal is to discuss the possibility of merging them together into just resources. This will end up with resource_types being defined as a resource in the pipeline and removing the resource_types section entirely. A resource type, in result, will just become a resource that will be used by another resource as its type (which will introduce recursive resources).
+
```
marco-m: I suggest to add a new section here that makes an example: an excerpt of the current configuration and how it would become with this proposal.

vito:
Chatted with @clarafu - gonna close this as things have evolved quite a bit in a new direction: #37 will replace `resource_types:` with `prototypes:` and we no longer plan to unify it with `resources:`.

PR #12, opened 2018-09-17
RFC: Storing users' team memberships

pivotal-jwinters:
[Proposal](https://github.com/pivotal-jwinters/rfcs/blob/proposal/group-storage/06-group-storage/proposal.md)

Status: Draft

This is a brain dump from a conversation we had about storing users' team memberships. This needs to be fleshed out a lot more, but it gives an early look at some of the options we've considered.

### DIFF
--- a/06-group-storage/proposal.md
+++ b/06-group-storage/proposal.md
@@ -21,7 +21,11 @@ Now that we've introduced `roles` it looks more like this:
 ```
 {
   "user_id": 1234567890,
-  "teams": ["main:owner", "some-team:viewer", "some-other-team:member"]
+  "teams": {
+    "main": ["owner"],
+    "some-team": ["viewer"],
+    "some-other-team": ["member"]
+  }
 }
 ```
 
--- a/06-group-storage/proposal.md
+++ b/06-group-storage/proposal.md
@@ -1,12 +1,32 @@
 # Summary
 
-The goal here is to start a discussion about how to store a user's team memberships. So far we've talked about two different options:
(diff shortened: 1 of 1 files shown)


Rukenshia:
@pivotal-jwinters Will this still synchronize with the login provider or would that mean that I would have to assign memberships within concourse? That would be a huge pain.

pivotal-jwinters:
@Rukenshia there's a few different ways we could approach this, but I don't think much would change from the operators perspective.

06-group-storage/proposal.md
```diff
@@ -0,0 +1,40 @@
...
+- It's operationally less complex. We could probably get rid of the skymarshal component altogether and rely directly on `dex` tokens.
+
+
+### Disadvantages
+
+- It's slower. Every api request requires a database lookup.
+
+- It's a more complex data model. More things to manage in the database.
+
+- We still don't have full control over upstream groups. This means that if you're added to a new team in github, for example, you would still need to log out and log back in for us to pick up those changes.
```
pivotal-jamie-klassen: Not sure this should count as a Disadvantage of storing team memberships in the database, as this problem exists for both strategies.

pivotal-jwinters: good point, i removed it

vito:
For anyone following along, this led to https://github.com/concourse/concourse/pull/4950 which has been merged and will be shipped in v6.1. Going to close this as there's not much point having an RFC for it retroactively.

PR #13, opened 2018-11-07
RFC: Spinnaker pipeline resource.

YoussB:
cc: @xtreme-sameer-vohra


xtreme-sameer-vohra:
[Proposal](https://github.com/YoussB/rfcs/blob/07-spinnaker-pipeline-resource/07-spinnaker-pipeline-resource/proposal.md)
Status: Draft

07-spinnaker-pipeline-resource/proposal.md
```diff
@@ -0,0 +1,130 @@
...
+
+The idea is build on an already existing [resource](https://github.com/burdzwastaken/concourse-spinnaker-resource).
+
+## Source Configuration
+
+- `spinnaker_api`: the url of the Spinnaker api microservice. 
+- `spinnaker_application`: The Spinnaker application you would like to trigger.
+- `spinnaker_pipeline`: The Spinnaker pipeline you would like to trigger.
+- `skip_ssl_verification`: **TODO** make sure it is possible.
+- `client_x509_cert`: Client [certificate](https://www.spinnaker.io/setup/security/authentication/x509/) to authenticate with Spinnaker.
```
burdzwastaken: below in the examples `spinnaker_x509_cert` is used. this should be consistent.

07-spinnaker-pipeline-resource/proposal.md
```diff
@@ -0,0 +1,130 @@
...
+- `client_x509_key`: Client [key](https://www.spinnaker.io/setup/security/authentication/x509/) to authenticate with Spinnaker.
```
burdzwastaken: below in the examples `spinnaker_x509_key` is used. this should be consistent.

07-spinnaker-pipeline-resource/proposal.md
```diff
@@ -0,0 +1,130 @@
...
+
+Places the following files in the destination:
+
+ - `metadata.json`: Contains the pipeline execution metadata returned from the Spinnaker [api](https://www.spinnaker.io/reference/api/docs.html#api-Pipelinecontroller-getPipelineUsingGET).
+
+ - `version`: A file containing the pipeline execution id.
+
+### `out`: Triggers a pipeline
+
+Triggers a Spinnaker pipeline.
```
burdzwastaken: `out` behaviour may also want to check for specific results within the pipeline before moving to the next phase. an example of this is:https://github.com/novas0x2a/concourse-jenkins-resource#out-trigger-a-build

YoussB:
Thanks @burdzwastaken for your comments. We have updated the RFC accordingly.

jchesterpivotal:
It is very helpful with certificates, tokens etc to have the choice between configuring them directly in params, passing them in environment variables or configuring a path that contains it.

This is somewhat true of other parameters, but certificates just seem to come up every time.

xtreme-sameer-vohra:
> environment variables

Hey @jchesterpivotal 
Did you have any specific use cases in mind or a reference resource that currently does this ?

If we add such functionality, it seems like the only way a user might be able to leverage it during the `check` stage would be to use their own custom image.

vito:
I don't think this should be merged into `concourse/rfcs` as this is not going to become part of the core distribution of Concourse. I think this is a reasonable process for collaborating on its design, but we should close this once you've all reached consensus.

xtreme-sameer-vohra:
Thats fair. We can just move the proposal to the README.md as it would serve as the documentation to use the resource.

vito:
Seems like this has run its course? Closing!

PR #14, opened 2018-11-15
RFC: Decouple API handlers from build execution

pivotal-jwinters:
[proposal](https://github.com/pivotal-jwinters/rfcs/blob/08-user-artifacts/08-user-artifacts/proposal.md)

Signed-off-by: Jamie Klassen <cklassen@pivotal.io>


08-user-artifacts/proposal.md
```diff
@@ -0,0 +1,74 @@
...
+As illustrated in the diagram below, the last four steps overlap concurrently.
+
+## what to do?
+
+Create an artifactserver, handling the following endpoints:
+
+1. `POST /api/v1/teams/:team/user-artifacts` to upload a folder to a volume
+1. `GET /api/v1/teams/:team/user-artifacts/:uuid` to download the contents of a
+   volume
+
```
topherbullock: These endpoints could also be under /api/v1/teams/:team/volumes where a POST can create a specific type of volume for the artifacts, and return the created handle. There could be other methods on individual volumes toGET the contents of the volume, and query parameters to find specific user artifacts.

What do you think?

We'll also want a separate interface which decouples this volume management behaviour from the internals of the worker and even the specifics of volumes and the pool of workers overall (per our meatspace conversation about the API handlers doing this directly). The K8s RFC kind of covers some of the details for that, and could maybe be changed to a generic "runtime interface" RFC.

pivotal-jwinters: The plan is to treat `artifacts`, similar to `resource_caches` and `task_caches`. So we're going to have another foreign key reference in the volumes table. Since the `volume` is the lower level object, we were thinking it made more sense to have an `artifacts` endpoint that implicitly creates a volume record (instead of the other way around). No real strong opinions here though. Maybe we could have a `volumes` endpoint that creates different types of volumes (resource_caches, task_caches, user_artifacts, etc..).

vito:
This RFC predates the RFC process and was more of a brain dump for internal refactoring, so I'm gonna close it (after confirming with @pivotal-jwinters).

PR #15, opened 2018-11-15
RFC: Create a scan queue for resource checking

pivotal-jwinters:
[proposal](https://github.com/pivotal-jwinters/rfcs/blob/09-scan-queue/09-scan-queue/proposal.md)

Signed-off-by: Jamie Klassen <cklassen@pivotal.io>

### DIFF
--- a/09-scan-queue/proposal.md
+++ b/09-scan-queue/proposal.md
@@ -1,0 +1,50 @@
+# scan queue
+
+Let's increase the symmetry between the radar and scheduler packages.
+
+Scheduling consists of
+1. creating pending builds (BuildScheduler)
+1. starting builds (BuildStarter)
+
+On the other hand, scanning consists of
+1. scanning/checking resources synchronously on a schedule or on demand
+   (resourceScanner.scan)
+
+If we took the two-step approach used in build scheduling, we could schedule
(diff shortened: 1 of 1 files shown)


09-scan-queue/proposal.md
```diff
@@ -0,0 +1,50 @@
...
+* allows for aborting resource checks
+* removes duplication in resource checks
+* allows for resource check history, and resource check error history
+
+## cons
+
+* two steps instead of one does introduce more complexity
+* we would need to decide if the check handler would poll for a result in the
+  database or if requests to the check endpoint would be async and anyone that
+  requires a response would need to poll a separate endpoint.
```
timrchavez: I vote for async check endpoint. If the time it took to do version checking could be bounded to some reasonable # or if we were only talking about one client (e.g. fly), I might be more in favor of waiting for a response since it's simpler, but as it is, I don't think it can be bounded short of picking some arbitrarily high timeout  (e.g. we're dealing with a 1.5GB mono repo that takes about 4-6m to fully clone onto a worker) in your client which could have you potentially trading one set of problems for another.

vito:
This has turned into real work and wasn't significantly user-facing so I think we'll just close it out. 👍

PR #16, opened 2018-12-12
RFC: Bootstrap RFC process

vito:
[Rendered](https://github.com/vito/rfcs/blob/master/README.md)

This is largely inspired by [Rust's RFC process](https://github.com/rust-lang/rfcs).

This also establishes early that all RFCs submitted to this repo will be licensed under Apache 2.


vito:
For posterity: this was mostly reviewed in person and I haven't heard any objections. We can always amend this with later RFCs. 🙂

Issue #5, opened 2018-06-08
(RF)RFCs: next iteration of credential manager support


vito:
I'm opening this issue as a call to action for anyone familiar with any of the existing credential managers:

* Vault
* CredHub
* AWS SSM
* AWS SecretsManager
* Kubernetes

I think it makes sense to discuss each of these as separate RFCs as I don't expect anyone is intimately familiar with *all* of them. If you're willing to throw your hat in the ring, go ahead and submit a PR with your own RFC! If there's only one specific aspect you feel that you're knowledgeable enough to propose, go ahead and start there. We can work together towards a consistent solution over time. We just need to get the ball rolling first.

# Feedback

We've already received a fair amount of feedback for some credential managers. Mostly Vault, as that's the first one we shipped and seems to have had the most traction (or at least the most opinionated people giving feedback!)

## Using arbitrary paths

* https://github.com/concourse/concourse/issues/1391
* https://github.com/concourse/concourse/issues/1814
* https://github.com/concourse/atc/pull/273

CredHub/Vault (probably all) users want to be able to use credentials at arbitrary paths, not just the Concourse-enforced schema (`/concourse/TEAM/PIPELINE/CRED`). This would allow secrets to be reused rather than copied.

Being able to use arbitrary paths would also allow Vault users to use different secret backends for each credential. The current implementation forces users to use the "generic" backend, if they're using recent releases of Vault.

This would be problematic at the moment because credential managers are globally configured, with a single auth scheme (not per-team). So permitting arbitrary path usage would result in team A being able to read team B's credentials, which is Very Bad. Which leads us to...

## Per-team credential manager configuration

* https://github.com/concourse/concourse/issues/1814

This would allow teams to use their own private credential managers, and allow one team to use CredHub while another uses Vault. This would allow teams to have different tokens with e.g. Vault, and delegate credential access to Vault's access controls rather than enforcing the path scheme.

We should be careful to encrypt the credential manager config now that it's sitting in the database.

## Credential caching

* https://github.com/concourse/atc/pull/236
* https://github.com/concourse/atc/pull/238

Right now Concourse resolves credentials every time they're used, which can hit Vault and CredHub pretty hard. @rfliam submitted a PR to cache Vault credentials, but in principle this is something we may be able to do generically. (Vault makes it easier to do correctly by having lease durations, but other backends could have some static configuration to determine this at least.)

## Other feedback?

[...]

PR #17, opened 2018-12-26
RFC: Separate secrets from pipelines

elgohr:
Merry christmas :-)

### DIFF
--- a/001-separate-secrets/proposal.md
+++ b/001-separate-secrets/proposal.md
@@ -1,0 +1,23 @@
+# Summary
+
+> Separate secrets from pipeline definitions
+
+# Motivation
+
+> At the moment secrets and pipelines are mixed together. You can insert a secret via a ((variable)).  
+> In this way you have to set and know every secret on every set-pipeline. There are ways to simplify this, like having it in a file and passing it into the set-pipeline; this makes it much more insecure, as you have to store the file somewhere and/or care for not commiting it to git.  
+
+# Proposal
+
+> At my point of view it would be nice to separate secrets from pipeline definitions. If there would be a way to set these separately, the following scenarios would be much easier to implement:
+* Update pipelines without even having the need to know the secrets.  
+This would be a big plus, for pipeline maintenance. Today people have to look up the secrets via get-pipeline and transfer it to the updated set-pipeline by hand. This is a big vector for errors.
+* Setting/Updating pipelines dynamically without manual interaction.  
+It would open up the door for having something like a ".travis"-file for concourse  
+>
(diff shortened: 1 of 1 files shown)


elgohr:
Hey, thanks for the hint.
This is nearly what I was looking for; I had a little more focus on pipelines, instead of ATC, in my mind.

Especially in big environments you have the need to put more and restrictions in terms of "who's allowed to see which secret?" into the providers configuration. Furthermore centralization was never good for secrets.

At my point of view this could be improved by having the ability to set one provider per team and/or making the providers transparent for the Concourse user by configuring it through the Concourse API. I would prefer the second variant.

001-separate-secrets/proposal.md
```diff
@@ -0,0 +1,23 @@
+# Summary
+
+> Separate secrets from pipeline definitions
+
+# Motivation
+
+> At the moment secrets and pipelines are mixed together. You can insert a secret via a ((variable)).  
+> In this way you have to set and know every secret on every set-pipeline. There are ways to simplify this, like having it in a file and passing it into the set-pipeline; this makes it much more insecure, as you have to store the file somewhere and/or care for not commiting it to git.  
+
+# Proposal
```
vito: In terms of what this proposes so far I too am a bit confused as it sounds like today's credential management already exists in service of the same goal. There seems to be a lot of overlap between https://github.com/concourse/rfcs/pull/17#issuecomment-450441398 and things being discussed in https://github.com/concourse/rfcs/issues/5 but this RFC doesn't mention any of that.

Maybe this RFC could be steered in a more precise direction based on feedback in #5? Right now it doesn't appear to be proposing any concrete changes and seems to ignore the existing credential manager support in general.

elgohr:
Thank you for mentioning https://github.com/concourse/rfcs/issues/5 . This is going into the direction what I was thinking. I'll close this RFC in favor of discussions in the other RFC

vito:
@elgohr RFCs may be proposed in an early stage so that the community can figure out the details, but they are meant to be centered around changes to propose so that the discussion is more focused and the impact of those changes can be more easily measured. It's difficult to know how to engage if it's too open-ended - reviewers may not know what direction it's trying to go.

RFCs are the step between issues and pull requests:

* 💭 Issue (sharing ideas, developing understanding) ->
* 💡 RFC (a concrete proposal has emerged, let's dig into the details and discuss impact/etc) ->
* ✈️ Pull request (a proposal has been implemented)

I think there are a few things mentioned in #5 that could have proposals made for them (probably as separate RFCs as the domain is quite large). But there hasn't been a *whole* lot of discussion there so that's why I suggested there may be more to do while still in the 'issues' phase, at least for some of the things mentioned (some of it is integration-specific, i.e. Vault).

But if there *is* something that can be proposed concretely that would also be a good start. For example, it feels like we're all in agreement that having per-team credential manager config would be better - the next step for that idea may be an RFC proposing what that may look like. 🙂

Issue #7, opened 2018-07-19
Reusable tasks


vito:
(this is a rough draft/placeholder just to centralize brainstorming before an actual RFC)

ref. concourse/concourse#417
ref. concourse/docker-image-resource#185

There is clearly a need to make tasks just as easy to share, version, and re-use just as resources are.

What would this look like?

Today you can approximate this by storing task `.yml`s in a Git repo, and then pulling them in as a resource.

One big hole is configuration. You can only really specify `params`, which for tasks are environment variables. Lots of resources use `params` for arbitrary, sometimes nested config. Should we extend tasks to have the same ability? It would actually be pretty easy to just pass it to the script on `stdin`, just like we do for resources. (Would we then rename today's `params` to `env` or something to be more clear?)

jchesterpivotal:
A few months backed [I sketched some notes for these](https://docs.google.com/document/d/1hszTDI1FpZfMeycDsTk3cXQ-w4DVATXjPPUYhlxBcGw/edit?usp=sharing). It goes a bit further than what's immediately needed -- I assume that streaming inputs are a supported case.

vito:
I wanna pull https://github.com/concourse/concourse/issues/608 into the discussion here too. I think if we had more shareable tasks we'd be more inclined to implement things like the Terraform resource as a task or set of tasks instead of a resource, and then the user is free to use whatever storage backend the like for persisting the task's side effects.

I recognise that this goes against the idea of tasks being pure and side-effect-free, but I think the more important aspect has always been that they're *idempotent*, and I'm interested in seeing if this change in mindset leads to a more expressive reconfiguration of existing ideas.

I've mocked up a pipeline for this here: https://gist.github.com/vito/3f0ed09648bf117d95a3b7daa99f61c2

Here are some decisions I made with this mockup:

* Tasks are defined at the top-level, and are referenced by name when used in jobs. Now there's an actual use for the name of the task step! That feels nice an intuitive - task names have been completely cosmetic for a while, unlike their `get`/`put` counterparts.
* Top-level tasks are provided by a resource, similar to resource types. This supports storing them in `git` repos pretty well. This pattern is already followed by [cf-deployment-concourse-tasks](https://github.com/cloudfoundry/cf-deployment-concourse-tasks) for example. I like storing tasks in `git` repos as it makes versioning and inspection very easy (browse around on GitHub rather than reverse-engineering a `Dockerfile`).
[...]

PR #19, opened 2019-01-21
RFC: Simple pipeline merging and templating

pms1969:


cppforlife:
@pms1969 just to throw in an alternative solution. @cdutra and i recently were discussing how one might use [ytt](https://get-ytt.io) to help manage complexity of concourse pipelines. given that one can define functions in ytt it becomes pretty easy to decompose and reuse jobs, tasks, etc. this i think nicely keeps concourse out of business of templating and allows other tools to do what they do best. (`ytt template -R -f . | fly -t prod set-pipeline --pipeline example --config -`)

pms1969:
@cppforlife ytt looks interesting.  I'll have a deeper look some time soon.

samrees:
I would not be a big fan of concourse bringing in the templating hell of helm-esque systems.
Maybe useful to mention this for anyone else seeking composability at a pipeline level: https://cuelang.org/docs/about/

ringods:
Why not drop any type of declarative language and adopt the solution [Pulumi](https://www.pulumi.com/) brings to Infrastructure as Code? You could even adopt Pulumi by implementing a resource provider against the currently unofficial REST api of Concourse.

vito:
Hi @pms1969 - sorry it took so long to get to this. We recently adopted a more fluid and transparent process for RFC review, so we're catching up on the backlog of RFCs. Thanks for putting this together!

My gut reaction is to stay out of the business of templating, since a) no one can agree on what templating system to use because everyone uses something different, and b) the need for templating is to me an indicator that there may be something missing in Concourse's design. IMO templating is a last resort; introducing it as a first-class feature before exhausting the possibilities for core improvements may introduce a local maxima that increases the cognitive overhead for users while reducing critical feedback loops that allow us to improve Concourse further.

In addition, I think a few of the items on the v10 roadmap address at least some of the underlying needs:

* #31 allows you to codify your templating + configuration flow; a preceding step can render the pipeline YAML before passing it to `set_pipeline` step
  * this won't help with `fly`, but the overall goal is to be using `fly` less and have your Concourse self-configure from a bootstrap pipeline (or project with #32)
* #32 allows you to split pipeline definitions up and share things like resource definitions between pipelines in the project's file tree
* #29 + #31 allows you to dynamically configure pipelines across either dynamic or static var matrixes which may be an alternative to templating multiple job definitions
  * #32 also aims to allow `passed` constraints to span pipelines, though it doesn't cover that in great detail at the moment (TODO added)
* #37 addresses #7 a bit more directly; instead of reusable tasks you'd just write prototypes, as you would with resource types

Given all that, I'm going to propose that we postpone this until more of the v10 roadmap is delivered. Per the RFC process this doesn't mean it's final say - it'll bump this RFC to the top of the table on `concourse-ci.org` to get more eyes on it, so it's a chance to get more feedback and for anyone to challenge the resolution.

PR #24, opened 2019-04-03
RFC: generalize resource interface

vito:
[Rendered](https://github.com/vito/rfcs/blob/generalized-resources/024-generalized-resources/proposal.md)

Status: **stable draft**

**Please leave comments on individual lines** so that conversations can be threaded and resolved. Comments left at the toplevel unnecessarily will be ~brutally destroyed~ maintained so that the conversation doesn't get out of hand.

Related RFCs which specify how this interface is interpreted:

* #26 (artifact resources)
* #27 (trigger resources)
* #28 (notification resources)
* #29 (spatial resources)

Previous discussions:

* https://github.com/concourse/concourse/issues/534
* https://github.com/concourse/rfcs/pull/1 - notably https://github.com/concourse/rfcs/pull/1#issuecomment-477749314 which has led to this new RFC


024-generalized-resources/proposal.md
```diff
@@ -0,0 +1,297 @@
...
+  Value string `json:"value"`
+}
+
+// TLSConfig captures common configuration for communicating with servers over TLS.
+type TLSConfig struct {
+  // An array of CA certificates to trust.
+  CAs []string `json:"ca_certs,omitempty"`
+
+  // Disable TLS, effectively making communication over TLS insecure.
+  InsecureSkipVerify bool `json:"insecure_skip_verify,omitempty"`
```
vito: This wording feels strange but I *think* I've seen this variation the most.

If I were to name this I would have called it `skip_verification`, but I guess the idea is to shove 'insecure' in there somewhere so people know not to use it. I'm not sure about 'skip_verify' vs 'skip_verificication' though - to me the latter feels more natural. :man_shrugging:

024-generalized-resources/proposal.md
```diff
@@ -0,0 +1,303 @@
...
+  // User-specified configuration.
+  Config Config `json:"config"`
+}
+
+// InfoResponse is the payload written to stdout from the `./info` script.
+type InfoResponse struct {
+  // The version of the resource interface that this resource type conforms to.
+  InterfaceVersion string `json:"interface_version"`
+
+  // An optional icon name to show to the user when viewing the resource.
```
vito: We should add namespacing to this per https://github.com/concourse/concourse/issues/3789

jchesterpivotal:
In terms of a very broad reaction, I'd like to revisit and repitch the concept that resources are _dimensions_ and that versions are _coordinates_, positions on a dimension. Any dimension is meant to be ordered, insofar as there is a rule for determining occurs-before and occurs-after for any given version.

At the moment dimensions are fixed at pipeline-submission time, rather than being able to flex. This was the motivation for spaces, but one issue was that we attempted to "nest" dimensions. There were first-order dimensions like `my-repo` and then second-order dimensions like `my-repo/pullrequest-27`.

In terms of versions, a _coordinate_ may point to something else (as in the Artifact resource case), or it may "point to itself", as in the notification and trigger cases. This means that we retain the rule that Concourse only ever deals in tracking versions or coordinates; if the coordinate is also self-describing, that's neither here nor there.

A remaining paradox is dealing with re-executions. Concourse only ever moves forward. Sometimes this causes problems when the resources are not designed to model a "move backwards" as being a compensating motion. For example, resources like `cf` have a forward deployment, but not a rollback. Adding a `delete` option is necessary for many cases, but is logically distinct from a rollback operation. In order to maintain the forward-only rule, the same resource either needs a version type that can represent rolling back, or a second resource with a distinct history.

A paradox that I don't think can be solved in the present model is simply re-running a job with the _same_ inputs. This is a frequent complaint, especially since resource versions may be consumed when the build merely errors, as opposed to failing. The ability to add `attempts` adds a way to  prevent some unwanted consumption. But it's not a "real" solution; it's arguably an orthogonal concern.

To improve the first paradox, and fix the second paradox, I'd suggest that Concourse adopts a bitemporal (really, bidimensional or tridimensional) concept of version/coordinate history. The first timeline already exists, which is the mapping of versions within a resource. This creates spans or intervals during which some condition (passed, failed, errored) is held to apply. This is elsewhere called valid time, event time or fact time. The word "time" is misleading, because in our case the ordering is created by git commits, file versions, BOSH deployments etc.

To enable re-building, a second dimension is added: a statement of when the fact was considered to be true (typically called transaction time, processing time or belief time). This is a dimension created by _builds_. The statements about the intervals in fact/valid/event time are bounded by the second timeline. This allows questions like "what did Concourse believe about commit abc123 _as of_ build 123?". And because it can safely encode as-of information, it becomes safe to re-build with the same resource versions, as many times as desired, with each build advancing the belief time by one increment.

This has been a fairly unstructured ramble. For more details I refer you to [another fairly unstructured ramble](https://docs.google.com/document/d/1ZlPp099_fV1lyYWACSyuWY_jfUUM4kTkhk_0AVejN1k/edit#heading=h.ylwqrf1avidp) where I explored some of these ideas previously.

jchesterpivotal:
Here's a second reply, having dug into the original linked comment some more, rather than just pattern-matching keywords.

I think the expanded config maps somewhat to what I think of as dimension coordinates. For something like `git`, the `source` block is a set of slowly-changing dimension coordinates (git repository, branch, username) and the `version` is the fast-changing dimension coordinates (git commits).

I don't feel I understand how resources get fully plugged together, though, especially for cases where notifications are attached to other resources to accumulate or augment their history.

vito:
@jchesterpivotal As mentioned in the OP, please comment on individual lines so conversations can be threaded. In this case you can just comment on some throwaway line if there's no particular line the response can be attributed to. I'm going to delete those soon as they're too open-ended to have at the top-level, but I'll give you time to migrate them.

jchesterpivotal:
No worries. I'll save off copies and put them into better shape.

024-generalized-resources/proposal.md
```diff
@@ -0,0 +1,316 @@
...
+
+* Standardize TLS configuration so every resource doesn't implement their own way: [concourse/rfcs#9](https://github.com/concourse/rfcs/issues/9)
+
+## Glossary
+
+* **Config**: an arbitrarily nested JSON object containing user-provided configuration
+  * Examples: `{"uri":"https://github.com/concourse/concourse"}`, `{"interval":"10m"}`
+
+* **Config fragment**: a smaller JSON object intended to be "spliced" into a **config** by assigning each field from the fragment into the config.
+  * Examples: `{"ref":"abcdef"}`, `{"branch":"develop"}`
```
vito: We should be careful with the allowed syntax for the keys - i.e. disallow `:` since we use it for version syntax in `fly check-resource`.

vito:
closing in favor of https://github.com/concourse/rfcs/pull/37

PR #25, opened 2019-04-18
RFC: decouple build history and scheduling

pivotal-jamie-klassen:
[proposal](./25-decouple-history-and-scheduling/proposal.md)


jchesterpivotal:
Some of this may be open to improvement using table partitioning, which has been progressively improving since landing in PostgreSQL v10.

clarafu:
I have definitely done some thinking around what to do with the gigantic builds table and I have thought of partitioning the builds table before. I looked into it a little and if we were to partition the builds table by each job, I remembered reading that the performance of it doesn't work really well when there are a lot of or a growing number of partitions. I also read that the later versions of Postgres like v12 has a lot of performance enhancements with large numbers of partitions (https://www.2ndquadrant.com/en/blog/postgresql-12-partitioning/) but we don't want to rely on users having the most updated version of Postgres. So Im siding on waiting until partitioning in Postgres is more stable since it seems like they are constantly working on improving it.

I also think this RFC is an interesting idea, I like the idea of having the old data be condensed so it doesn't need to take up all the rows in the database. A few thoughts I have about it:
- With build scheduling, we will walk over all the previous builds if necessary in order to find one that matches the passed constraints we are looking for. This means we will be looking at the `builds`, `build_resource_config_version_inputs`, `build_resource_config_version_outputs`, `build_pipes` and `successful_build_outputs` tables. Each row in these tables will be referencing a row in the `builds` table through the `build_id` foreign key with delete cascade, so if we remove the old build rows and condense them into one row we will need some way of condensing the rows or keeping them around in the other tables referencing the old builds.
- Would they be separate tables that would contain the old data? Since the columns would need to basically be specific to the condensed data. Also, if the data is condensed it will be hard to use postgres to query out the specific data that we need. For example, in the algorithm we query out any build that matches these versions. If we were to condense the rows into text, we will no longer be able to use postgres queries to grab the specific build.
- Within the algorithm I think the flow would be, try to find a build that matches my passed constraints in the `builds` table, if it is not found, look into the table with the old rows. We will need to parse out the condensed data and try to find a build that matches our constraints.
- For viewing the build in the web UI, I guess it would be similar to the algorithm where it will need to fetch builds from the `builds` table and if there are no more to fetch, try to grab them from the condensed old builds table. We will need some way to keep around old build events too because they have a foreign key to the `builds` table.
- We also have containers that can reference a `build_id`. We might have to think about all the cases where we ave a foreign key to `build_id`, that there won't be any case where we will try to insert a `build_id` that is "old" and is moved to the condensed old data.

jchesterpivotal:
> I looked into it a little and if we were to partition the builds table by each job, I remembered reading that the performance of it doesn't work really well when there are a lot of or a growing number of partitions.

FWIW, I'd probably pick dates as the partitioning key. Most builds are occurring at the edge of new versions. There should be very little activity on old build records, so shuffling them into partitions will reduce search time. The open question would be what range size to use, I presume we'd need some load testing to sniff out a sensible default.

As much as possible, I don't think we should change the schema into current-vs-old data. This really is what partitions are meant to ease. Even if we use very coarse partitions (say monthly), we could still see substantial speedups on ATCs with dense multi-year histories.

clarafu:
I see, that's actually a really good point of partitioning by date instead since old records will "most likely" not be touched often! I think that's something that I can look into, because this current vs old data seems like it does come with a lot of complications. I'll do a bit of research and update here if I think this should be viable right now. Maybe I'll create a new RFC for it too. If we do partition sparsely (like monthly), I don't think I would be super worried about a large number of partitions and the performance handicap it might have.

Thanks for the amazing input!! 😊

PR #26, opened 2019-05-21
RFC: artifact resources

vito:
[Rendered](https://github.com/vito/rfcs/blob/artifact-resources/026-artifact-resources/proposal.md)


vito:
closing in favor of https://github.com/concourse/rfcs/pull/38

PR #28, opened 2019-05-21
RFC: notification resources

vito:
[Rendered](https://github.com/vito/rfcs/blob/notification-resources/028-notification-resources/proposal.md)

### DIFF
--- a/028-notification-resources/proposal.md
+++ b/028-notification-resources/proposal.md
@@ -5,6 +5,7 @@ This proposal uses the [generalized resource interface](../024-generalized-resou
 ## Motivation
 
 * Support notifications in a way that doesn't pollute pipeline config and UI: [concourse/concourse#1052](https://github.com/concourse/concourse/issues/1052), [concourse/rfcs#10](https://github.com/concourse/rfcs/issues/10)
+* Support notifying on state change only (i.e. don't spam on 'fail', 'fail', 'fail').
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


kallisti5:
Feedback? One thing we *really* miss from buildbot is "State change only" notifications.
Example:
  * Success
  * Success
  * Success
  * Error!
  * Error!
  * Success

Instead of 6 notifications, having the option of only 2 would be a lot better from a "preventing spam" standpoint.  We rolled our own custom notification plugin, and still wasn't able to accomplish it. (no persistent data between notification plugin runs)

vito:
@kallisti5 Good call! I've listed that under 'Motivation' so I don't forget. :+1:

agonbina:
Maybe this is already addressed in the specification, but would also be interesting to version notifications so that you can update them to avoid too much spam (ex. a slack notification to notify build started, can then be updated to "failed/succeeded") etc.

Anonymous-Coward:
I was led here from [here](https://github.com/concourse/rfcs/discussions/107).

Noob question: why not replace a dedicated notifications resource with a dedicated events resource, with events being able to trigger arbitrary tasks/jobs? Would increase the flexibility and power of concourse overall, and bring it closer to its stated purpose of being an automation tool. 

The current approach, of triggering jobs only on get of resources is often not that convenient. Events, OTOH, are quite flexible.

kallisti5:
I think "onFailure" does exist for notifications now?  ... so this one might be a little less relevant than it once was.   State change only events would be a nice feature though.

Anonymous-Coward:
> I think "onFailure" does exist for notifications now? ... so this one might be a little less relevant than it once was. State change only events would be a nice feature though.

I was thinking of more generic events - maybe arbitrary identifiers you can attach to success, failure and finally - just like exception handling. Then, specify the event as a trigger for another job in the same pipeline.

taylorsilva:
Not doing this anymore.

PR #29, opened 2019-05-21
RFC: `across` step

vito:
[Rendered](https://github.com/vito/rfcs/blob/spatial-resources/029-across-step/proposal.md)


029-across-step/proposal.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Triggering on changes
+
+With `trigger: true` configured, the build plan will run on any change to the
+set of `((vars))` - i.e. when a var is added or removed.
+
+
+## Open Questions
+
+* It seems like triggering could work in both the static and dynamic cases, but
```
mxplusb: Assuming I understand the difference between static vs dynamic, I could see static triggers as empty payloads having value for things like smoke tests, where you want to continuously test your system, so you want a static trigger where the value of the trigger is irrelevant (i.e. time resource), you just need the trigger to have fired. So let's say that I have two identical (for the sake of argument) environments, beta and gamma, and I want to run scheduled smoke tests, I could see it being something like this:

```yaml
resources:
- name: 5m
  type: time
  source: {interval: 5m}

plan:
- across: [beta, gamma]
  as: 5m
  do:
  - task: smoke-tests
    file: smoke-tests-task.yml
```

Again, assuming I understand the semantics of dynamic vs static, that's where my thought process lead. In this case the value of any of the inputs is diminished because the task config is not needed due to some default input in the task, the value of `across` in this case is that the build matrix was executed statelessly. Does that makes sense?

mxplusb:
How would across work with a large build matrix? 

I could see the value here of `across` having some functional limiter, like this:

```yaml
across: [1.8, 1.9, 1.10, 1.11, 1.12, 1.13, 1.8-win, 1.9-win, 1.10-win, 1.11-win, 1.12-win, 1.13-win, 1.8-macos, 1.9-macos, 1.10-macos, 1.11-macos, 1.12-macos, 1.13-macos, etc.]
max_in_flight: 4
as: go-version
do:
- task: unit
  vars: {go_version: ((go_version))}
```

That way it limits the parallelism of the build matrix a bit without stealing too much from the global and job `max_in_flight` and prevent the pipeline from moving forward on other tasks in the same pipeline/job.

JohannesRudolph:
Is it planned that `across` also adds a notion for "skipping" a job for certain resource combinations? Say we have a build matrix but one particular job may not apply to a certain combination of vars, would it be possible to skip running the job? Or should that rather be implemented inside the job itself (i.e. NOP-ing all tasks inside)? I'm asking because it has proven very difficult so far to implement pipeline control flow, see e.g. https://discuss.concourse-ci.org/t/trigger-jobs-conditionally/306/13

I know that pipeline control flow is a controversial topic and having "idempotent" pipelines that always do the same thing is a good design goal for Concourse. However there's already `try` and `on_success` and `on_failure`. However, they don't support all real world use cases, especially when it comes to optimising pipelines. I'd rather have concourse support some sort of declarative notion (like `job.if` [in GitHub Actions](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif)) than abusing task scripts or other mechanisms.

029-across-step/proposal.md
```diff
@@ -0,0 +1,199 @@
...
+- var: pr
+  source: booklit-prs
+- var: go_version
+  values: [1.12, 1.13]
+```
+
+This will run 2 * (# of PRs) `set_pipeline` steps, setting two pipelines per
+PR: one for Go 1.12, and one for Go 1.13.
+
+### Controlling parallelism with `max_in_flight`
```
aoldershaw, line 82: The `in_parallel` step also has a `fail_fast` option - is this something that also makes sense for the across step?

Similarly, does it make sense to use the term `limit` rather than `max_in_flight` for consistency with `in_parallel`, or do you think that'd just lead to confusion?

vito: Good points!

I had the same debate over `max_in_flight` vs `limit`, but I think `max_in_flight` is less ambiguous in this specific context as  `limit` could also mean "limit to first N vars".

The `in_parallel` step is a bit inconsistent in other ways: we opted to go with a nested config format for its additional settings (`in_parallel: {steps: [...], limit: 3, fail_fast: true}`) to avoid polluting the `PlanConfig` type, but with the step refactor (https://github.com/concourse/concourse/pull/5504) it now has its own `InParallelStep` type, so that wouldn't be a problem anymore, and we could easily change it to this syntax:

```yaml
in_parallel:
- step1
- step2
fail_fast: true
limit: 3
```

I recall @itsdalmo (who PR'd the `in_parallel` step) suggesting it would be confusing to have the term `max_in_flight` in multiple places (on the [job](https://concourse-ci.org/jobs.html#schema.job.max_in_flight) and on the `in_parallel` step), but, well, they all mean the same thing, so it could end up helping folks understand and discover the option. So now I wonder if we should change the `in_parallel` step to `max_in_flight` and a flat config syntax. (It would be easy to do backwards-compatibly; just add more fields to the `InParallelStep` struct.)

Re: `fail_fast` - hmm, I think this definitely makes sense to add, but it would be hard to add in the current syntax, since it'd be an attribute of the whole `across` step modifier, and modifiers can't have additional sibling fields. I was able to get away with putting `max_in_flight` under the vars list but that trick wouldn't make sense for `fail_fast`.

I wonder if I should just change this back to being its own step rather than a step modifier. That would clear up precedence questions between `across` and `timeout`/`attempts`/`on_*`/`ensure` (i.e. whether they apply to the modified step or the original step).

Something like this:

```yaml
across:
- var: go_version
  values: [1.14, 1.15]
- var: postgres_version
  values: [9.5, 10]
each:
  task: unit
  vars:
    go_version: ((.:go_version))
    postgres_version: ((.:postgres_version))
fail_fast: true
```

The only part of this syntax I don't really like is that it raises a question of whether `each` is a list of steps or a single step. Making it a list would make it easier to run multiple steps, but it would be inconsistent with the `on_*` hooks. Keeping it a single step would be consistent with the hooks, but then you'd have to use `do:`. 🤷 

One alternative would be to do something like what `in_parallel` did, which is to support either a list of vars or a config, but, well, it feels kind of clunky:

```yaml
task: unit
vars:
  go_version: ((.:go_version))
  postgres_version: ((.:postgres_version))
across:
  vars:
  - var: go_version
    values: [1.14, 1.15]
  - var: postgres_version
    values: [9.5, 10]
  fail_fast: true
```

aoldershaw: If we decide to go back to `across` being its own step, what do you think of renaming `each` to `do` (similar to the previous revision of the RFC):

```yaml
across:
- var: go_version
  values: [1.14, 1.15]
- var: postgres_version
  values: [9.5, 10]
do:
- task: unit
  vars:
    go_version: ((.:go_version))
    postgres_version: ((.:postgres_version))
fail_fast: true
```

I think this would make it clear(er) that it's a list of steps rather than a single step (since it looks like a modifier that only applies to the `do` step - although maybe this could lead to some confusion where people would try to apply it to other steps as well?)

vito: @aoldershaw Yeah, that's the tricky part: if we overload `do:` it could end up confusing folks into attaching it to other step types, and it might be weird to see a "normal" `do:` after that.

vito: For us Elm fans... :P

```yaml
let:
- var: go_version
  values: [1.14, 1.15]
- var: postgres_version
  values: [9.5, 10]
in:
  task: unit
  vars:
    go_version: ((.:go_version))
    postgres_version: ((.:postgres_version))
fail_fast: true
```

vito: Hmm...

Here's a thought: if we make it possible for step modifiers to have sibling fields, we could support `fail_fast` in the modifier form and just use the `do:` step as a work-around when it's necessary to disambiguate (pretending `in_parallel` has it as a toplevel field for a moment).

And then if you just put the `do:` below the `across:`, it's exactly the syntax you proposed, but with no gotchas - it's actually a `do` step, and you could replace it with a single step if you wanted. Here's a full example showing its use for disambiguation:

```yaml
across:
- var: go_version
- var: postgres_version
do:
- in_parallel:
  - task: unit
    vars:
      go_version: ((.:go_version))
      postgres_version: ((.:postgres_version))
    on_failure: # send test diagnostics
  fail_fast: false # (or omitted)
on_failure: # send notification
fail_fast: true
```

Precedence would be `ensure` < `on_*` < `attempts` < `timeout` < `across` < (core steps). This would enable modifiers to be applied to the `across` step itself, with the `do:` syntax available for when you need to apply them (or conflicting fields like `fail_fast`) to the inner step.

The cool thing is that this syntax would transparently work while implemented as a modifier - it's just that you're modifying the `do:` step, and you've placed it below its modifier!

It might not be too difficult to support sibling fields - I think it'd just take a bit of `reflect` magic to remove all struct fields instead of just the one used to identify the step. 🤔

aoldershaw: Huh, that's pretty neat! FWIW I find the `across` syntax a bit easier to follow when written like your example above (with `across` as the first field in the step) rather than when written at the end like other modifiers

aoldershaw: I can see this possibly being a bit of a gotcha, but what if we didn't remove the sibling `fail_fast` field for `across`, so writing:

```yaml
across:
- var: go_version
- var: postgres_version
in_parallel:
  - task: unit
    vars:
      go_version: ((.:go_version))
      postgres_version: ((.:postgres_version))
  - ...
fail_fast: true
```

would apply the `fail_fast` both to the `across` step **and** to the `in_parallel` step (the assumption being that if you wanted to fail fast, you'd usually want to fail fast everywhere - which may be a wrong assumption). In the case where you didn't want the `across` step or the `in_parallel` step to fail fast, you'd wrap the `in_parallel` in a `do`

vito: I've updated the proposal to include `fail_fast` and document the step precedence plus the `do:` syntax workaround.

I ended up going with this precedence order instead: `ensure` < `on_*` < `across` < `attempts` < `timeout`

The reason for this is so that `attempts` is attached to the inner step and not the entire matrix. It doesn't seem like anyone would ever want to rerun the entire matrix in the event that one step fails.

`timeout` is also important to have applied to the inner step: if it were applied to the matrix, `max_in_flight` throttling could easily cause the timeout to be exceeded before some steps even run.

029-across-step/proposal.md
```diff
@@ -0,0 +1,277 @@
...
+plan:
+- set_pipeline: pr
+  instance_vars: {pr_number: ((.:pr.number))}
+  across:
+  - var: pr
+    source: booklit-prs
+```
+
+The above example will run the `set_pipeline` step across the set of all GitHub
+PRs, returned through a `list` operation on the `booklit-prs` var source.
```
aoldershaw, line 61: This may not be relevant discussion to this RFC, but I'm a bit confused about how `var_sources` work in the context of PRs. It seems to me like there's a separation between credential manager `var_sources` and PR-like `var_sources`.

It don't really see the use case to run a plan `across` a list of credentials from Vault, nor do I see the use case to access a single PR via a `var_source` (e.g. `((booklit-prs:1234))` to get some info about PR#1234? - based on your comment here https://github.com/concourse/rfcs/pull/37#discussion_r428195212, it sounds like you'd want to `get` a PR via the resource interface instead)

vito: The unifying theory between var sources is only this: they discover a set of things (credential paths, branches, PRs) and have a way to fetch additional data for each thing (credential value, branch/PR metadata). It's not really suggesting that they all have the same usage - only that they fit the same shape, and Concourse has primitives that work with those shapes.

Who knows: maybe someone does have a use case for using `across` with a credential manager? Perhaps they want to monitor all credentials to detect whether any of them have been accidentally checked in to the codebase, or whether they are due to be rotated.

It's true that you would use a `get` step to fetch the PR's bits, but you may also want to get the attributes of a PR in order to, say, periodically check that the PR assignees have been active. This could be done through a job in the PR pipeline that fetches the PR var:

```yaml
jobs:
- name: ping-assignees
  plan:
  - get_var: time
    source: daily
    trigger: true
  - get_var: pr
    path: ((pr_number))
    source: booklit-prs
  - run: check-activity
    type: pr-monitor
    params:
      assignees: ((.:pr.assignees))
      last_activity_at: ((.:pr.last_activity_at))
      last_active_user: ((.:pr.last_active_user))
```

note: I cheated there - `get_var` doesn't have a `path:` - but it probably should, otherwise you'd have wacky syntax like `((.:((pr_number)).assignees))`.

aoldershaw: Makes sense, thanks for the clarification! Those are some pretty cool use cases, and as you say, I'm sure there's plenty more people will come up with.

Excited for more of these v10 ideas to land!

aoldershaw: Sorry, one last thought around this - given that secrets from credential managers aren't all necessarily the same shape, do you think it makes sense to extend the var interpolation syntax to provide a way to coerce any var into a string? For instance:

```yaml
task: detect-leaked-secret
vars:
  secret: ((.:secret | json)) # (or `| yaml`)
across:
- var: secret
  source: vault
```

Since otherwise the job would error if any secret isn't a string

vito: @aoldershaw Hmm, it's true that with `vars:` you would likely need some sort of marshalling because even though `vars:` supports complex structures, the task config template wouldn't really know what to do with it. I think you could avoid marshalling/unmarshalling values if you use a Prototype and pass the value along through `params:` instead, where it would become part of the JSON request structure. I would almost prefer this approach so that folks write more prototypes instead of tasks.

Also, one tricky detail with credential scanning: `across` would get the vars from `list`, which would not include the sensitive values themselves. I think credential managers should omit sensitive values from the `list` call so that we can safely list available credentials in the web UI (to authorized parties, anyway). Doing so would really help with discoverability: "what was that credential called again?" -> navigate to the var source -> use `(())` syntax. So you would need a `get_var` step somewhere to fetch the actual credential.

029-across-step/proposal.md
```diff
@@ -0,0 +1,277 @@
...
+values to set when executing the step. The step will be executed across all
+combinations of var values.
+
+### Static var values
+
+The set of values can be static, as below:
+
+```yaml
+task: unit
+vars: {go_version: ((.:go_version))}
```
aoldershaw, line 30: This is more of a question about the var steps RFC, but since that's merged, I'll put my question here.

The ["var steps" RFC](https://github.com/concourse/rfcs/blob/master/027-var-steps/proposal.md#proposal) states:

> Each build will have a single local scope which steps can use for setting and getting var values

Does this mean that the var fetched from a `load_var` (or `get_var`) step nested under e.g. a `do` block will be accessible outside of that `do` block?

```yaml
plan:
- do:
  - load_var: my-var
    file: some-file.yml
- task:
  vars:
    my-var: ((.:my-var)) # Is this allowed? My guess is yes, but just checking
```

If so, how does this work with the `across` step? Presumably, each "iteration" of the base step would need its own local scope (at least for the vars that get iterated over):

```yaml
plan:
- across:
  - var: version
    values: [v1, v2, v3]
    max_in_flight: all
  do:
  - task: some-task
    vars:
      version: ((.:version)) # ((.:version)) has to take on 3 different values, which suggests multiple scopes
```

Additionally, are the var steps allowed to be used from inside the across step? (borrowing the modified `get_var` syntax from https://github.com/concourse/rfcs/pull/29#discussion_r452312783)

```yaml
plan:
- across:
  - var: pr
    source: booklit-prs
    max_in_flight: all
  do:
  - get_var: pr-details
    path: ((.:pr.number))
    source: booklit-prs
  - task: some-task
    vars:
      assignees: ((.:pr-details.assignees))
- task: another-task
  vars:
    assignees: ((.:pr-details.assignees)) # presumably, pr-details would not be in scope here
```

vito: Haha, yep, good catch. I think we'll need some sort of scoping mechanism for the `across` step to have any hope of functioning in a sane way.

I think the `across` step should perhaps run its sub-steps with their own scope that inherits from the outer one and has their vars set there. That way they can still reference outer vars, and their own vars, and any vars they set within the step (like your last example) would only be reachable from within their scope - i.e. they won't "leak" to the outside.

I'll add a section for this! The var steps RFC doesn't really cover it because there isn't any notion of sub-scopes at the moment - this would be the first instance of it.

029-across-step/proposal.md
vito: @mxplusb Sorry, didn't see this by the time I got back to this RFC!

I think your example above would be written like this:

```yaml
resources:
- name: 5m
  type: time
  source: {interval: 5m}

plan:
- get_var: 5m
  trigger: true
- across:
  - var: env
    values: [beta, gamma]
  do:
  - task: smoke-tests
    file: smoke-tests-task.yml
```

This question was more about whether it makes sense to configure `trigger: true` on the `var: env` part. The reason it's an interesting case is that the values will only ever change when the config changes, so it's a little weird to 'trigger' on it, but I decided we should support it since it still makes sense from the perspective of 'I changed the config, and I want a change to this config to auto-trigger a new build'.

re: your other comment (`max_in_flight`), that has been added to the proposal. :)

029-across-step/proposal.md
```diff
@@ -0,0 +1,321 @@
...
+  config:
+    repository: vito/booklit
+    access_token: # ...
+
+plan:
+- set_pipeline: pr
+  instance_vars: {pr_number: ((.:pr.number))}
+  across:
+  - var: pr
+    source: booklit-prs
```
aoldershaw, line 57: Have you thought about allowing the `across` step to iterate over objects from an arbitrary prototype (i.e. not just constrained to `var_source` prototypes and their `list` message)? That could make the object streams returned by prototypes more useful outside of the special prototype interfaces that Concourse will interact with natively (resources, var_sources, notifications, etc.)

In this comment https://github.com/concourse/concourse/pull/5936#issuecomment-667192099, you suggested that the `run` step could possibly support a `set_var` field to store the returned object as a local variable. Perhaps an alternative to `set_var` would be to use the `across` step in its place, e.g.

```yaml
- across:
  - var: some-var
    type: gval
    params: {fields: {some-field: hello}}
  run: echo
  type: debug
  params: {message: ((.:some-var.some-field))}
```

Kind of weird in this case since we're iterating over a single value always - but I guess for `set_var` to work in general, it would also have to set the var to a list of objects (even if a single object is emitted)

aoldershaw: This also makes me realize that the `across` step can be used for conditional portions of build plans, by having the list of values be either 0 or 1 element long 🤔

aoldershaw: It's also making me wonder about dynamic build plans in the context of the `across` step (i.e. the list of values to iterate over can be determined at runtime) - what are your thoughts on that? I think it could greatly extend the use cases that `across` can support, but it could possibly make pipelines difficult to understand since you could introduce arbitrarily complicated control flow with the point above

vito: I have, and your comments are pretty much the same thought process I went thought. 😄 It seems like a really cool idea, but yeah, the difficulty is that it seems to require some sort of dynamic build plan construct.

I'm running into another area that would benefit from dynamic build plans: the way `file:` and `image_resource:` interact on a task step. I'm working on a refactor that pulls image fetching out into explicit `check` and `get` steps in the build plan, but the existence of `file:` means I can't know whether they're needed (Darwin and Windows have no images, for example), or what image to even `check` and `get`, until the build is running. In this case I can probably just preemptively configure steps which may just do no-ops, and they can probably get their config in a similar way to how the `Get` plan uses `VersionFrom` to get some config from a prior step, but it all feels kind of held together with duct tape. (There's also a bit of complexity in that we may need to `check` + `get` a custom type that the `image_resource:` uses, but I think have a way to get around that.)

I still don't see a super obvious way to implement dynamic build plans though. 🤔 But I haven't really tried, either, I guess.

aoldershaw: The only thought I had was for there to be some sort of build event that says "inject this build plan under this plan ID" or "replace this plan ID with this build plan". So, the plan would initially look something like:

```json
{
  "schema": "exec.v2",
  "plan": {
    "id": "across-id",
    "across": {}
  }
}
```

and there could be a build event like:

```json
{
  "event":"dynamic-build-plan",
  "version":"1.0",
  "data": {
    "id": "across-id",
    "across": {
      "some": "data"
    }
  }
}
```

which the UI would interpret to construct a final build plan of:

```json
{
  "schema": "exec.v2",
  "plan": {
    "id": "across-id",
    "across": {
      "some": "data"
    }
  }
}
```

I'm sure there are some technical challenges I haven't considered, though

vito: Ooh, I like the idea of emitting plans in the event stream - seems a lot simpler than e.g. modifying and re-saving the build plan directly.

It'd be a little interesting to figure out the best way to structure it, for example how to indicate how plan A is related to plan B so that the UI can know how these dynamic plans should be displayed. For my `task` step example from earlier, there may be N plans if we e.g. need to perform `check` and `get` of a full hierarchy of chained resource types leading up to the one used in `image_resource`. Maybe these could all be given some sort of 'image-fetching' tag or something. 🤔

aoldershaw: Just trying to understand the use case you're describing/the refactor you're working on - is it that whenever we use an image (be it in a `get`, `put`, `task`, or `check`) we would have a `check` + `get` for the (potential chain of) parent `resource_types` and for the anonymous `image_resource` for `task` steps - but the latter can only be determined at runtime?

Does it make sense for that "image fetching" plan to be a part of the plan for those step types? For instance, a `put` step could have a plan like this:

```json
{
  "id": "1",
  "put": {
    "name": "some-resource",
    "type": "some-resource-type",
    "image_fetching_plan": {
      "id": "2",
      "do": [
        {
          "id": "3",
          "check": {"type": "registry-image"}
        },
        {
          "id": "4",
          "get": {"type": "registry-image"}
        },
        {
          "id": "5",
          "check": {"type": "some-resource-type"}
        },
        {
          "id": "6",
          "get": {"type": "some-resource-type"}
        }
      ]
    }
  }
}
```

(I simplified what the plan would look like for the explicit `get` and `check` steps, since I'm not sure what they'll look like exactly. Also, does it apply to base resource types like `registry-image` as well?)

---

For `task` steps with a `file:`, we could initialize the `task` plan to:

```json
{
  "id": "1",
  "task": {
    "name": "some-task",
    "image_fetching_plan": {
      "id": "2",
      "do": []
    }
  }
}
```

and after we have the `file:` artifact, we can send a `dynamic-build-plan` event:

```json
{
  "event":"dynamic-build-plan",
  "version":"1.0",
  "data": {
    "id": "2",
    "do": [
        {
          "id": "3",
          "check": {"type": "registry-image"}
        },
        {
          "id": "4",
          "get": {"type": "registry-image"}
        },
        {
          "id": "5",
          "check": {"name": "some-image"}
        },
        {
          "id": "6",
          "get": {"name": "some-image"}
        }
    ]
  }
}
```

Does that make any sense?

vito: @aoldershaw Yeah that makes a lot of sense.

My current approach is as follows:

* Use `CheckPlan` to run a `check` for the image
* Use `VersionFrom` on the `GetPlan` to get the version returned from the `CheckPlan`
  * Normally `VersionFrom` this is used for the `get` after a `put`, but it seems to be a clean fit here too
* Add `ImageArtifactName` to all other core plans (`check`, `get`, `put`), and use to refer to the artifact fetched by the `GetPlan`
  * Normally `ImageArtifactName` is used by the `TaskPlan` for when [`image:`](https://concourse-ci.org/jobs.html#schema.step.task-step.image) is set on a `task` step.
* Remove all image resource `check`/`get` logic from `atc/worker/`, now that everything is either a base resource type or an already-fetched image artifact

So far I've just chained it all together using `OnSuccessPlan` like this:

```json
{
  "id": "1",
  "on_success": {
    "step": {
      "id": "check-type-id",
      "check": {
        "type": "registry-image",
        "source": {
          "repository":"concourse/git-resource"
        }
      }
    },
    "hook": {
      "id": "3",
      "on_success": {
        "step": {
          "id": "4",
          "get": {
            "name": "type:my-git",
            "type": "registry-image",
            "source": {
              "repository":"concourse/git-resource"
            },
            "version_from": "check-type-id"
          }
        },
        "hook": {
          "id": "5",
          "put": {
            "image_artifact_name": "type:my-git",
            "source": {
              "uri": "https://github.com/vito/booklit"
            }
          }
        }
      }
    }
  }
}
```

The downside of this approach is that I think the UI will literally render this as a bunch of `OnSuccess` hooks, which isn't ideal. I could use `do`, but that'll also look surprisingly different from what the user expects. So some approach like yours, which retains the original plan structure and instead adds the setup plans as sub-fields of the steps which need it, could be a better structure.

I haven't actually gotten this code to work yet - I immediately ran into a bit of a roadblock, as it's impossible to inter-mix `CheckPlan` with any other kind of plan. `CheckPlan`s can only be run as LIDAR checks, cannot be run within builds, and other types of plans cannot be added to LIDAR checking plans - it nil panics if the toplevel plan isn't a `CheckPlan`.

So now I'm going down a rabbit hole of having all `check`s run as builds instead of having their own separate stack. The idea is to end up with two build schedulers: a resource scheduler alongside the job scheduler. So far it looks promising; it feels kind of nice to have builds be the common substrate for everything that we run - it may be the same way that we approach Projects, which also execute build plans.

If you want to take a peek at what I have so far, the branch is [`checks-as-builds`](https://github.com/concourse/concourse/tree/checks-as-builds). So far it's super spiky - once I have more confidence in this being a good path to go down I plan to break it up into many smaller changes and be a bit more strategic about it.

029-across-step/proposal.md
```diff
@@ -0,0 +1,321 @@
...
+WARNING: across step shadows local var 'foo'
+one
+two
+zero
+```
+
+
+## Open Questions
+
+* n/a
```
vito, line 290: @JohannesRudolph Sorry, completely forgot to address your point! (Raising this as a review comment so the replies can be threaded.)

Would it be sufficient to allow a static list of combinations to skip to be provided?

```yaml
across:
- var: go
  values: [1.16, 1.15, 1.8]
- var: platform
  values: [darwin, linux, windows]
except:
- platform: windows
  go: 1.18
```

This would work similarly to [excluding jobs in Travis CI](https://docs.travis-ci.com/user/build-matrix/#excluding-jobs), i.e. a subset of vars may be specified, and all variations with those vars will be skipped.

aoldershaw: Related thought: it would be kind of cool if we could `except` combinations based on a relationship between variables. e.g. if you want to test various upgrade paths, you only want to test combinations of `from` and `to` where `to > from`. I have no idea how this would work - it might just be a use case for var sources rather than `except`

029-across-step/proposal.md
vito: Follow-up: the dynamic sub-steps stuff was delivered in v7, which now runs `check`s as builds and also performs image fetching as sub- `check` and `get` steps! 🎉

We're using more of this dynamic sub-step stuff for https://github.com/concourse/concourse/issues/6274

Given this new capability it's actually pretty tempting to re-frame this as something like this:

```yml
- get_vars: foos
  trigger: true
- across:
  - var: foo
    values: ((.:foos))
  do: # ...
```

It's a bit more verbose, but I like that it lets you trigger off of changes to the set without having to `across:` something. This could be used for re-building a website based on the set of PRs, for example, instead of having to iterate over them in the build plan. (I have this use case right now while I'm working on adding an RFC status board to the website. 😄)

...actually, what if we just made the `across:` value be a list of steps to run, and then we go over all of their results? We could make `values:` be an actual step that runs and returns the predefined list of values, and then you could do `run:` or `get_vars:` or whatever you want too.

```yml
across:
- var: foo
  get_vars: foos
  trigger: true
- var: foo
  values: [1, 2, 3]
- var: foo
  run: blah
```

aoldershaw: I like that idea! Small suggestion - what if instead of inlining in the var definition, we supported `values` and `values_from`, e.g.:

```yaml
across:
- var: foo
  values_from:
    get_vars: foos
    trigger: true
- var: bar
  values: [1, 2, 3]
```

I feel like that's maybe a little more obvious/readable at a glance.

Also, I feel like the name `get_vars` is a little awkward, especially because (and correct me if I'm wrong) the `list` operation of `var_sources` doesn't necessarily give the "values" of the vars (it'd be a bit risky for a `vault` var_source to potentially leak all credentials, even if you have access to them). Maybe it could be something like `list_vars`?

---

Here's a bit of a wild idea that poses several complications (and I don't even think is a good idea really, kinda just spitballing) - but what if `values_from` could be any step that produced a stream of values. This could be `get_vars`, `load_var`, `run`, or a `do` where the last step produces a stream. Then, you could do stuff like:

```yaml
- get: ci
- across:
  - var: foo
    values_from:
      do:
      - get: data
        trigger: true
      - task: filter-list
        file: ci/tasks/filter-list.yml
        input_mapping: {input: data}
        output_mapping: {output: filtered}
      - load_var: filtered
        file: filtered/data.json
```

Which would then imply 2+ levels of checking/triggering - one level triggers on changes to the `trigger` resources/vars (which runs a `check` or `get_var`), and on changes to the resource/var, we trigger a build that runs the full `do` plan + any `get` steps in advance (perhaps we require inputs to be explicitly specified in the case of a `task`). e.g. in this case, the plan would be something like:

```yaml
- get: ci
- get: data
- task: filter-list
  ...
- load_var: filtered
  ...
```

...and when `filtered` changes, we trigger the rest of the parent build plan

Again, don't think this is really a good idea haha, just throwing it out there

vito: That seems close to what I'm proposing - whatever the step results in gets used as the value. I think it'd just look like this with the syntax I had before:

```yaml
var: foo
do:
- get: data
  trigger: true
- task: filter-list
  file: ci/tasks/filter-list.yml
  input_mapping: {input: data}
  output_mapping: {output: filtered}
- load_var: filtered
  file: filtered/data.json
```

...but this is just a different syntax. `values_from` is probably clearer. I think we'd still need to handle `values: ((.:foo))`, but that could probably be handled by the `across` step too instead of a separate `values:` step.

In either case, I'm not sure how the `across` step would access the result in the case of a `do`. 🤔 Currently results are stored under step IDs stored in the `RunState`, but this is almost like a return value from an expression. With a single step it's obvious what step ID to use, but with a `do` it's not so easy. It's almost like we need `(exec.Step).Run` to be able to return a value on its own, and `do` would return the last value. But that changes the interface.

Maybe this whole storing-results-in-`RunState` pattern could/should change? @clarafu @chenbh Has your recent experience in this area led to a similar line of thinking?

vito: Oh, also re: `get_vars` vs `list_vars`, yeah depending on what the resulting value is `list_vars` might be more accurate. I think either `get_vars` would actually fetch all the var values individually, using a `list` call under the hood for the scheduling/triggering, or we would have `list_vars` instead and allow the user to run a separate `get_var` step for each var value if that's desired. Don't know which; haven't put much thought into it yet.

aoldershaw: What I was proposing is a bit more complicated than just a multi-step `values_from` (where the value comes from the last step), but I didn't have the proper framing. Now, I realize it the proposal was actually a combination of two ideas:

1. A new way to communicate between jobs by emitting/receiving data
2. Anonymous jobs (run nested in other jobs)

I'll flesh out my original example a bit more:

```yaml
jobs:
- name: parent
  plan:
  - get: ci
  - across:
    - var: foo
      values_from:
        do:  # <----------------------------- A
        - get: data
          trigger: true <-------------------- B
        - task: filter-list
          file: ci/tasks/filter-list.yml
          input_mapping: {input: data}
          output_mapping: {output: filtered}
        - load_var: filtered # <------------- C
          file: filtered/data.json
    ...
```

Here, A is an "anonymous job". When the trigger on B gets a new version, the `parent` job doesn't trigger - only the anonymous job A. The `load_var` C implicitly emits the value it loads, and `parent` implicitly receives and triggers on changes to this value.

The implicit emitting and receiving is very non-obvious, though, and the anonymous job idea is a bit odd (either we make it fully self-contained and need to duplicate the `get: ci`, or it magically can borrow artifacts from the parent job). What if instead we made it explicit and gave the filtering it's own job:

```yaml
jobs:
- name: filter-list
  plan:
  - get: ci
  - get: data
    trigger: true
  - task: filter-list
    file: ci/tasks/filter-list.yml
    input_mapping: {input: data}
    output_mapping: {output: filtered}
  - load_var: filtered
    file: filtered/data.json
    emit: data # <--------------------- A

- name: parent
  plan:
  - get: ci
  - receive: data <-------------------- B
    job: filter-list
    trigger: true
  - across:
    - var: foo
      values: ((.:data))
    ...
```

A emits a new value and stores it as `data`. B triggers on changes to that emitted value and stores the result in a local variable.

This mechanism could even be used for `{get_var: ..., trigger: true}` - the builds we run periodically to fetch var could `emit` the value to the DB, and the job would `receive` that changed value from the build.

chenbh: > In either case, I'm not sure how the across step would access the result in the case of a do. 🤔 Currently results are stored under step IDs stored in the RunState, but this is almost like a return value from an expression. With a single step it's obvious what step ID to use, but with a do it's not so easy. It's almost like we need (exec.Step).Run to be able to return a value on its own, and do would return the last value. But that changes the interface.

I thought of it like every type of step would have a single output (e.g. `check` is version, `get` is image spec, `load_var` and `get_var` are var), in `do`'s case maybe we just need to figure out the contract for what it should return (is it time to introduce a `result` step/modifier)? 

It would then be up to `across` to figure out what to do with this info (which might be wild cause that implies we can run `across` a list of versions from a `check`).

kurtmc: I am not sure if I am adding anything useful to the conversation, but when I saw the the 7.0.0 release notes I immediately saw a good use case for one of our pipelines and I tried to build a working prototype using pipeline instances and the across step.

```yaml
- name: set-pipeline-instances
  serial: true
  plan:
    - in_parallel:
      - get: ci
      - get: src-branches
        trigger: true
    ...
    - load_var: branches
      file: branches-json/branches.json
    - set_pipeline: {{ (datasource "config").name }}
      file: ci/ci/build.yml
      instance_vars: {branch: ((.:b))}
      across:
        - var: b
          values: ((.:branches))
```

I am keen for the normal resource type to trigger the job and then use the `load_var` to load a variable that is a list and then just use that for for the argument to `values`.

aoldershaw:
> would there be a way to rerun only the failed combinations without having to rerun all the combinations again?

@drunkirishcoder that's discussed [here](https://github.com/vito/rfcs/blob/spatial-resources/029-across-step/proposal.md#modifier-syntax-precedence). It's just a matter of adding an `attempts` modifier to the step:

```yaml
task: unit
timeout: 1h # interrupt the task after 1 hour
attempts: 3 # attempt the task 3 times
across:
- var: go_version
  values: [1.12, 1.13]
on_failure: # do something after all steps complete and at least one failed
```

drunkirishcoder:
@aoldershaw cool. that can auto retry x attempts. but would there be a manual way? like in the concourse UI rerun the task with the same settings but only retry the failed combo?

029-across-step/proposal.md
```diff
@@ -0,0 +1,321 @@
...
+Other examples of modifiers are `timeout`, `attempts`, `ensure`, and the `on_*`
+family of hooks.
+
+In terms of precedence, `across` would bind more tightly than `ensure` and
+`on_*` hooks, but less tightly than `across` and `timeout`.
+
+`ensure` and `on_*` bind to the `across` step so that they may be run after the
+full matrix completes.
+
+`attempts` binds to the inner step because it doesn't seem to make a whole lot
```
aoldershaw, line 178: @drunkirishcoder going to move this discussion to a review comment thread to avoid cluttering the top-level.

> that can auto retry x attempts. but would there be a manual way? like in the concourse UI rerun the task with the same settings but only retry the failed combo?
>
> the use case I'm trying to design for is I want to use `across` to deploy to multiple clusters/environments. and if one of the env failed, I wouldn't want to rerun the deployment to the successful ones again. so ideally if I can rerun the same matrix, but only for the failed ones again.


Ah, so that's what you meant. There are no plans for this currently, and it would be a bit challenging to get right - there's no precedence for partial reruns of a build. In general, we'd still need to run everything before the `across` step again, since those steps may produce artifacts that the `across` step relies on (e.g. `get`s or `task` outputs).

I think the ideal solution would for the deploy steps to be idempotent - i.e. given the same inputs, if the deployment already succeeded the step would no-op, so re-running the whole matrix wouldn't be an issue. That's perhaps easier said than done, though.

Is there a reason why manual retries is preferable to using `attempts` for automated retries?

drunkirishcoder: @aoldershaw ok, thank you for the detailed response. that makes a lot of sense. as to why prefer manual over automated retries. I'm just thinking sometimes a deployment would fail due to circumstances that may be longer to recover, like someone forgot to apply a firewall rule. but yeah I understand why it would be difficult to do in concourse. thanks.

vito: One thing you could do is configure a pipeline for each cell in the matrix using the `set_pipeline` step, rather than setting up all the environments within the single build. That way you could retrigger the failed job in each pipeline.

029-across-step/proposal.md
aoldershaw, line 290: Something that didn't really come up in the proposal is how we handle steps that emit artifacts (e.g. `get`, `task.outputs`, prototype outputs). Currently, the implementation of the static `across` step runs each iteration in its own local artifact scope. This means that any artifacts go out of scope past the `across` step and cannot be used.

It'd be cool if there was a way to "fan-in" on artifacts emitted within an `across` step. For instance, suppose there was a `prototype` (or just a `task`) for compiling a Go binary that outputs an artifact called `binary`. You might want to compile it for multiple platforms/architectures, which could make use of the `across` step:

```yaml
across:
- var: platform
  values: [linux, darwin, windows]
- var: arch
  values: [amd64, arm64]
run: compile
type: go
params:
  package: repo/cmd/my-program
  platform: ((.:platform))
  arch: ((.:arch))
outputs: [binary]
```

(side-note: it might make more sense from a performance PoV for a `go` prototype to support compiling multiple platforms/architectures in a single `run` step where possible, but bear with me for the example)

The trouble is - under the current implementation, there's no way to access `binary` outside of the `across` step. Even if we used a shared artifact scope for each iteration, it'd just be "last write wins" (so, with no parallelism, `binary` would be the result for `(windows, arm64)` only) - since they all share a name, the results would clobber each other. This could be avoided by encoding more information in the output name via an `output_mapping` (e.g. name the artifacts `binary-linux-amd64`, `binary-linux-arm64`, ...) - but that only works for simple values like strings/numbers.

So, what if we let artifacts in an `across` step share a name, but be uniquely identified by the set of vars that were used to produce that artifact (in the same way that instanced pipelines share a name and are uniquely identified by their `instance_vars`). A possible syntax for referencing the binary created for `(linux, amd64)` could be `binary{platform: linux, arch: amd64}`

Here, `binary` really refers to a matrix of outputs (that mirrors the `across` matrix). What if we also provided a way to filter down that matrix to get at a subset of the outputs. e.g. `binary{platform: windows}` would give the (2) outputs built for windows, and `binary{arch: arm64}` would give the (3) outputs built for arm64.

So, suppose you wanted to upload a github release containing *all* 6 binaries produced in the matrix. You could do something like:

```yaml
put: release
params:
  globs: [binary/*] # or maybe [binary{*}/*] to make it clear it's a matrix
```

Suppose you wanted to bundle the different architectures separately - then, you could do:

```yaml
across:
- var: arch
  values: [amd64, arm64]
put: ((.:arch))-bundle
params:
  globs: [binary{arch: ((.:arch))}/*]
```

This may warrant a separate proposal but figured I'd bring it up here since it's definitely a limitation of the `across` step as it stands

dumez-k:
Hi @taylorsilva have there been any updates to this epic recently? Is this still under active development? 

Also a little confusing from looking through the initial [spatial resources RFC](https://github.com/concourse/concourse/issues/1707) but was support ever implemented for this in any release version? Or is all of this work just prototype unreleased stuff?

Thanks!

029-across-step/proposal.md
multimac: One idea, maybe we could re-use the concept of `output_mappings`. Similar to how a task defined in a file can have its outputs mapped by an `output_mappings` configuration in the pipeline referencing that task, an `across` step could have its outputs made available via a similar mapping.

As an example:
```yaml
across:
- var: arch
  values: [amd64, arm64]
- var: version
  values: [1.0, 1.1]

task: create-artifact
output_mapping:
  output: task-artifact

across_output_mapping:
- vars: # match vars from an execution of the task
    arch: amd64
    version: 1.0
  outputs: # map an output from the modified step to an output of the across step
    task-artifact: amd64-bundle-v1.0
```

This would be a fairly verbose if you wanted to map every output from the matrix of steps:
```yaml
# [...]
across_output_mapping:
- vars: {arch: amd64, version: 1.0}
  outputs: {task-artifact: amd64-bundle-v1.0}
- vars: {arch: arm64, version: 1.0}
  outputs: {task-artifact: arm64-bundle-v1.0}
- vars: {arch: amd64, version: 1.1}
  outputs: {task-artifact: amd64-bundle-v1.1}
- vars: {arch: arm64, version: 1.1}
  outputs: {task-artifact: arm64-bundle-v1.1}
```

Though it would possibly simpler to implement, without relying on prototypes or other Concourse functionality.

When, or if, [this limitation](https://concourse-ci.org/across-step.html#limitations) on using vars in `get:` and `put:` steps (eg. `get: bundle-((.:arch))-((.:version))`) is lifted, the `across_output_mapping` could be updated to allow for the use of vars as well, simplifying the configuration to something like...

```yaml
# [...]
across_output_mapping:
- outputs:
    task-artifact: ((.:arch))-bundle-v((.:version))
```

taylorsilva:
Merging this now. Not everything has been implemented but it's in a "good enough" state now that people can already use it for the original purpose.

The only things missing are triggering on `var_sources` changes, which actually depends on `var_sources` being further developed.

PR #31, opened 2019-07-11
RFC: `set_pipeline` step

vito:
[Rendered](https://github.com/vito/rfcs/blob/set-pipeline-step/031-set-pipeline-step/proposal.md)

### DIFF
--- a/031-set-pipeline-step/proposal.md
+++ b/031-set-pipeline-step/proposal.md
@@ -40,7 +40,7 @@ The pipeline would be automatically unpaused, as opposed to `fly set-pipeline` w
 
 ## `((vars))` support
 
-Additionally, we should support `vars` (as in `fly set-pipeline -y`) and `vars_files` (i.e. `fly set-pipeline -l`):
+Additionally, we should support `vars` (as in `fly set-pipeline -y`) and `var_files` (i.e. `fly set-pipeline -l`):
 
 ```yaml
 plan:
@@ -48,7 +48,7 @@ plan:
 - set_pipeline: release
   file: ci/pipelines/release.yml
   vars: {release_version: 5.3}
-  vars_files:
+  var_files:
   - ci/pipelines/vars/foo.yml
 ```
 
--- a/031-set-pipeline-step/proposal.md
+++ b/031-set-pipeline-step/proposal.md
@@ -36,7 +36,9 @@ The `x` in `set_pipeline: x` is the pipeline name, and `file:` would be used to 
 
 The pipeline would be configured within whichever team the build execution belongs to.
 
-The pipeline would be automatically unpaused, as opposed to `fly set-pipeline` which pauses pipelines by default. The assumption here is that if you're automating `set_pipeline` you're not just kicking the tires and can probably trust the pipelines that you're configuring are correct, at least enough to have made it into version control.
+Upon first configuration pipeline will be automatically unpaused, as opposed to `fly set-pipeline` which puts newly configured pipelines in paused state by default. The assumption here is that if you're automating `set_pipeline` you're not just kicking the tires and can probably trust the pipelines that you're configuring are correct, at least enough to have made it into version control.
+
+When configuring an existing pipeline, however, the pipeline's paused status will not be changed. In other words, the `set_pipeline` step will leave already-existing paused pipelines in the paused state. The assumption here is that the pipeline has been manually paused by a pipeline operator, possibly in response to an emergent situation, and it should be left alone.
 
 ## `((vars))` support
 
--- a/031-set-pipeline-step/proposal.md
+++ b/031-set-pipeline-step/proposal.md
@@ -54,7 +54,34 @@ plan:
 
 # Open Questions
 
-none
+* `set_pipeline: self`
+
+  Currently, the `foo` in `set_pipeline: foo` is the name of a pipeline to set.
+  A pipeline could technically update itself by configuring its own name in the
+  step, but pipeline configs aren't meant to contain their own name, as doing
+  so prevents the config from being re-used as a 'pipeline template'.
+
+  Are self-updating pipelines a feature that we want to explicitly support by
+  allowing the keyword `self` in place of the pipeline name?
+
+  The [Projects RFC][projects-rfc] outlines a more 'git-ops' style flow for
+  configuring pipelines, where instead of having pipelines self-update they are
+  all configured in one central place (the project's `plan:`).
+
+  Pros:
+
+  * Fairly straightforward semantics which seem to support a natural follow-up
+    question after learning about the `set_pipeline` step.
+  * Keyword use has precedent in `version: every`/`version: latest`/`inputs: all`.
+
+  Cons:
+
+  * Supporting both self-updating pipelines and projects could cause confusion
+    and fragmentation; it doesn't seem wise to have two competing approaches to
+    the same goal.
+  * Given that `self` isn't *critical* (it's easy to work around through
+    templating, i.e. `set_pipeline: ((name))`), is it worth the
+    risk/maintenance?
 
 
 # Answered Questions
@@ -65,6 +92,7 @@ none
 # New Implications
 
 ## Deprecating `concourse-pipeline` resource
+
 Deprecating the `concourse-pipeline` resource should be the primary goal.
 
 Some of the extended functionality of the resource will not be supported in the name of keeping the `set_pipeline` step design simple and easy to reason about.
@@ -72,3 +100,6 @@ Some of the extended functionality of the resource will not be supported in the 
 For example, the step should only ever configure one pipeline at a time - it should not support the `pipelines:` functionality for configuring a bunch at once.
 
 Similarly, the step should not support fully dynamic configuration (`pipelines_file:`).
+
+
+[projects-rfc]: https://github.com/concourse/rfcs/pull/32
--- a/031-set-pipeline-step/proposal.md
+++ b/031-set-pipeline-step/proposal.md
@@ -61,6 +61,14 @@ plan:
 
   Or should we emit a warning?
 
+* Should we support glob expansion in `var_files`?
+
+  The `concourse-pipeline` resource supports this by just performing glob
+  expansion against its local filesystem. For the `set_pipeline` step, this is
+  a bit more challenging - there *is* no local filesystem. Would we have to
+  implement glob expansion in the Baggageclaim API or something? How easily
+  would this translate to other runtimes?
+
 * `set_pipeline: self`
 
   Currently, the `foo` in `set_pipeline: foo` is the name of a pipeline to set.
--- a/031-set-pipeline-step/proposal.md
+++ b/031-set-pipeline-step/proposal.md
@@ -1,0 +1,74 @@
+# Summary
+
+This RFC proposes a new `set_pipeline` step type for configuring pipelines within a build plan.
+
+
+# Motivation
+
+## Short-term motivation
+
+Lots of folks are already using the [`concourse-pipeline` resource](https://github.com/concourse/concourse-pipeline-resource), however the resource has two fatal flaws:
+
+* Users have to configure a local auth user and pass it to the resource definition.
+* The resource is versioned independently of the user's Concourse, meaning the `fly` version won't always be in sync. The resource makes an attempt to resolve this by doing a `sync` after logging in, but this is a pretty clunky problem regardless.
+
+If we had native support for a `set_pipeline` step, both of these problems would go away.
+
+## Long-term motivation
+
+By having a `set_pipeline` step in the build plan, we can start to improve Concourse's story around automating the full CI stack for projects of all sizes. Users can start to trust that pipelines are always configured via CI, and they can go over the build history to see who changed what and when.
+
+Later RFCs (namely, 'projects' and 'instanced pipelines') will build on this idea to provide a truly continuous workflow for automating pipelines - including their automatic archival when they're no longer needed, in the case of instanced pipelines.
+
+
+# Proposal
+
+Using the step would look something like this:
+
+```yaml
+plan:
+- get: ci
+- set_pipeline: concourse
+  file: ci/pipelines/concourse.yml
+```
+
+The `x` in `set_pipeline: x` is the pipeline name, and `file:` would be used to specify the pipeline config.
+
+The pipeline would be configured within whichever team the build execution belongs to.
+
+The pipeline would be automatically unpaused, as opposed to `fly set-pipeline` which pauses pipelines by default. The assumption here is that if you're automating `set_pipeline` you're not just kicking the tires and can probably trust the pipelines that you're configuring are correct, at least enough to have made it into version control.
+
+## `((vars))` support
+
+Additionally, we should support `vars` (as in `fly set-pipeline -y`) and `vars_files` (i.e. `fly set-pipeline -l`):
+
+```yaml
+plan:
+- get: ci
+- set_pipeline: release
+  file: ci/pipelines/release.yml
+  vars: {release_version: 5.3}
+  vars_files:
+  - ci/pipelines/vars/foo.yml
+```
+
+# Open Questions
+
+none
+
+
+# Answered Questions
+
+none
+
+
+# New Implications
+
+## Deprecating `concourse-pipeline` resource
+Deprecating the `concourse-pipeline` resource should be the primary goal.
+
+Some of the extended functionality of the resource will not be supported in the name of keeping the `set_pipeline` step design simple and easy to reason about.
+
+For example, the step should only ever configure one pipeline at a time - it should not support the `pipelines:` functionality for configuring a bunch at once.
+
+Similarly, the step should not support fully dynamic configuration (`pipelines_file:`).
--- a/031-set-pipeline-step/proposal.md
+++ b/031-set-pipeline-step/proposal.md
@@ -54,54 +54,117 @@ plan:
   - ci/pipelines/vars/foo.yml
 ```
 
+## Preventing manual updates
+
+When using `fly set-pipeline` to update a pipeline that has been configured
+through the `set_pipeline` step, a warning will be printed and a confirmation
+dialogue will be presented.
+
+When configured through `fly set-pipeline` thereafter, warnings will no
+longer be issued.
+
+This is to prevent accidentally configuring changes that will be blown away,
+while still allowing pipeline operators to take over its configuration if
+needed.
+
+
+# Experiments
+
+There are a few extended pieces of functionality that have been proposed. There
+is currently no consensus on these being the ideal long-term design, because
+there are alternative methods we're planning that should make them unnecessary.
+
+However, there is value in supporting them "until we get there." We can
+implement support for them, and include a warning both in their usage and in
+the documentation that they may be removed in the future.
+
+Each experiment must have an easy-to-find GitHub Discussion so that we can
+collect feedback on how the feature is used and confirm that the long-term
+design addresses the core need appropriately.
+
+## `set_pipeline: self`
+
+* PR: [#4857](https://github.com/concourse/concourse/pull/4857)
+* Feedback: [#5732](https://github.com/concourse/concourse/discussions/5732)
+
+Currently, the `foo` in `set_pipeline: foo` is the name of a pipeline to set. A
+pipeline could technically update itself by configuring its own name in the
+step, but pipeline configs aren't meant to contain their own name, as doing so
+prevents the config from being re-used as a 'pipeline template'. You could of
+course turn this into a var, but that's a little clunky to use.
+
+To support self-updating pipelines without making them self-aware, we can allow
+the keyword `self` to mean the current pipeline. There is precedent for such a
+keyword in other fields like `version: every`, `version: latest`, `inputs:
+all`, and `inputs: detect`.
+
+One downside of this approach is it doesn't cover the full lifecycle of the
+pipeline: who set it initially, so that the `set_pipeline: self` step can even
+run?
+
+This is a question that will likely be answered by the [Projects
+concept][projects-rfc] once it's introduced. Projects are designed to be the
+authoritative source for pipeline configuration, covering both the initial
+creation and the later updating of all pipelines contained therein.
+
+As such, it will be a little odd to support both `set_pipeline: self` and
+Projects side-by-side. But until Projects lands, there is benefit in allowing
+it so that we can confirm that Projects covers all the use cases for it by
+analyzing user feedback.
+
+## Setting pipelines in other teams
+
+* PR: [#5729](https://github.com/concourse/concourse/pull/5729)
+* Feedback: [#5731](https://github.com/concourse/concourse/discussions/5731)
+
+The `set_pipeline` step is designed to be a "piece of the puzzle" - just like
+other steps like `get`, `put`, and `task`.
+
+It is designed to operate against *one* pipeline, in the current team, and in
+the current Concourse cluster. This is in contrast to the
+[`concourse-pipeline` resource][concourse-pipeline-resource], which supports
+setting *many* pipelines across *many* teams within *any* Concourse cluster.
+
+This step is not intended to be a drop-in replacement for the
+`concourse-pipeline` resource, but it *is* a goal to deprecate it. However
+full deprecation is blocked on further development around the [Projects
+concept][projects-rfc] or other ideas that lead towards auto-configuring the
+full Concourse cluster.
+
+The `concourse-pipeline` resource provides significant enough burden to
+maintainers and users that it is probably wise to expedite its deprecation
+without waiting on these farther-off goals. To this end, we can
+experimentally support setting pipelines in other teams by configuring a
+`team:` field on the step:
+
+```yml
+set_pipeline: foo
+team: bar
+file: ci/foo.yml
+```
+
+This must only work if the step is being run by an admin team (i.e. `main`),
+making its usage somewhat limited. Once a more suitable replacement arrives
+this field can be removed.
+
+
 # Open Questions
 
-* Should we prevent pipelines configured via a `set_pipeline` step from being
-  updated manually through `fly set-pipeline`?
-
-  Or should we emit a warning?
-
-* Should we support glob expansion in `var_files`?
-
-  The `concourse-pipeline` resource supports this by just performing glob
-  expansion against its local filesystem. For the `set_pipeline` step, this is
-  a bit more challenging - there *is* no local filesystem. Would we have to
-  implement glob expansion in the Baggageclaim API or something? How easily
-  would this translate to other runtimes?
-
-* `set_pipeline: self`
-
-  Currently, the `foo` in `set_pipeline: foo` is the name of a pipeline to set.
-  A pipeline could technically update itself by configuring its own name in the
-  step, but pipeline configs aren't meant to contain their own name, as doing
-  so prevents the config from being re-used as a 'pipeline template'.
-
-  Are self-updating pipelines a feature that we want to explicitly support by
-  allowing the keyword `self` in place of the pipeline name?
-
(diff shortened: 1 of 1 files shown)


031-set-pipeline-step/proposal.md
```diff
@@ -0,0 +1,74 @@
...
+- get: ci
+- set_pipeline: concourse
+  file: ci/pipelines/concourse.yml
+```
+
+The `x` in `set_pipeline: x` is the pipeline name, and `file:` would be used to specify the pipeline config.
+
+The pipeline would be configured within whichever team the build execution belongs to.
+
+The pipeline would be automatically unpaused, as opposed to `fly set-pipeline` which pauses pipelines by default. The assumption here is that if you're automating `set_pipeline` you're not just kicking the tires and can probably trust the pipelines that you're configuring are correct, at least enough to have made it into version control.
```
Rukenshia: We store our ci pipeline along with our code in the source code repository, in a `ci/` directory. Would I integrate the `set_pipeline` step as the first step after the `get:` of my source code? Would changing the pipeline mid-run do anything to the active execution (as the `version` of my source code wouldn't change, but I am doing a set-pipeline)?

031-set-pipeline-step/proposal.md
```diff
@@ -0,0 +1,74 @@
...
+# New Implications
+
+## Deprecating `concourse-pipeline` resource
+Deprecating the `concourse-pipeline` resource should be the primary goal.
+
+Some of the extended functionality of the resource will not be supported in the name of keeping the `set_pipeline` step design simple and easy to reason about.
+
+For example, the step should only ever configure one pipeline at a time - it should not support the `pipelines:` functionality for configuring a bunch at once.
+
+Similarly, the step should not support fully dynamic configuration (`pipelines_file:`).
```
ari-becker: We're heavy users of `pipelines_file`.

We use the [github-list-repos resource](https://github.com/coralogix/eng-concourse-resource-github-list-repos) to dynamically fetch a list of repositories which need pipelines set for them, generates pipelines for them, and sets the resulting pipelines. As a benefit, we can create new services as necessary, and pipelines are automatically created, maintained, and consistent with each other, for every single one of our services. Currently this requires the use of `pipelines_file` because the list of services themselves is changing and dynamic.

031-set-pipeline-step/proposal.md
```diff
@@ -0,0 +1,74 @@
...
+
+The `x` in `set_pipeline: x` is the pipeline name, and `file:` would be used to specify the pipeline config.
+
+The pipeline would be configured within whichever team the build execution belongs to.
+
+The pipeline would be automatically unpaused, as opposed to `fly set-pipeline` which pauses pipelines by default. The assumption here is that if you're automating `set_pipeline` you're not just kicking the tires and can probably trust the pipelines that you're configuring are correct, at least enough to have made it into version control.
+
+## `((vars))` support
+
+Additionally, we should support `vars` (as in `fly set-pipeline -y`) and `vars_files` (i.e. `fly set-pipeline -l`):
```
DanielJonesEB: I think this is the most appropriate place to make this comment. I am simple and this is complicated, so please bear with me as I try to explain.

Have you considered use cases where the `set-pipeline` step needs to _use_ vars from the `vars_files`?

We chain together pipelines (all from the same `pipeline.yml` from the same repo, albeit with different versions) that represent different environments (ie dev, staging, prod). The `get`s of important resources that we want to test and promote together have parameterised `version` fields.

1. In the 'dev' pipeline, `get: { version:` is `latest` for all important resources we're integrating, **including the CI resource** that the pipeline is set from
1. Dev pipeline runs, integration tests run.
1. At the end of the pipeline, we use [stopover](https://github.com/EngineerBetter/stopover) to emit a YAML file of the versions of all resources we want to promote into the next environment.
1. Versions YAML file is uploaded to S3
1. The 'staging' pipeline was already configured to look at the S3 bucket, and thus a run of the pipeline is triggered. Notably, at this stage **the pipeline is configured according to an earlier version of the CI resource** (the previous one that passed 'dev').
1. The `set-pipeline` job runs, using the versions file as a `vars_file`. Because all the `get`s' `version`s are parameterised, this effectively constrains versions of resources available to this pipeline. At the point `set-pipeline` is invoked the runtime state of the pipeline is such that **the version of the CI resource is still set to that of the last run**, because the new vars haven't been applied - that won't happen until _after_ `set-pipeline` has succeeded.
1. The runtime state of the pipeline is now that the `get` of the CI resource is constrained to the newer version that passed 'dev'. 
1. We then need to run `set-pipeline` _again_, because only _now_ is the newer version that passed `dev` available to this pipeline
1. Pipeline runs as normal
1. Repeat until production glory

We use this pattern with multiple customers for their pretty important infrastructure stuff. In case anyone's wondering, the benefit it is *guarantees* that no untested versions can sneak into higher environments.

So I guess I have two questions, and because I'm not very clever the answer might already be clear based on the RFC, so apologies if you have to spell this out:

1. **Is the approach above compatible with the new proposal?** My understanding is that is _should_ be, as long as there's no limit on where the new `set_pipeline` step can appear, and how often.
1. **Could the new functionality do this atomically**? Being able to avoid the `set-pipeline` 'double-tap' would make things a lot more clear for us and our customers.

You have my contact details @vito, so let me know if I can help by explaining in more detail on a call or similar.

031-set-pipeline-step/proposal.md
aegershman: Would the proposed [`across`](https://github.com/concourse/rfcs/pull/29/files#diff-faa5c2911815511474d027ac56212456R15) semantic work for this?

ari-becker: > ^Would the proposed [`across`](https://github.com/concourse/rfcs/pull/29/files#diff-faa5c2911815511474d027ac56212456R15) semantic work for this?

@aegershman Yes and no? Different pipelines also need to be set per team. The `set_pipeline` RFC so far indicates that it should only be able to set pipelines for the team currently running the job, but our meta pipeline is on a special `meta` team which sets pipelines for other teams, and currently the decision of which team to set a pipeline for is also dynamic.

031-set-pipeline-step/proposal.md
vito: Yep, it would typically follow a `get` step today. The semantics of changing the pipeline mid-run are the same semantics as with `fly set-pipeline` today.

Once Projects come along (#32) we can eliminate that `get` step and start providing better guarantees about keeping task code in sync with pipeline versions and such.

031-set-pipeline-step/proposal.md
vito: FWIW, I don't think we can easily add support for setting pipelines for other teams. That re-introduces a bit of an auth predicament.

With `set_pipeline` acting on its already-owning team, you at least have the guarantee that someone with authority to set *its* pipeline would also have had the authority to set other pipelines within the team.

Once you can specify other teams, that guarantee is no longer there and we'd need some other construct for determining it. There's no team hierarchy right now - users are just authorized via each team's configuration, and that notion of which user set the pipeline isn't kept (and would be a little awkward IMO - now you have pipelines with associated individuals).

Aside from that, the `across` step feels appropriate for setting a pipeline for each repository - you would point it to a resource to detect the repositories themselves.

I'm definitely still interested in chatting about the across-teams use case but it may be wise to leave out of the first iteration. I suspect once we have all these tools in place (including projects, #32) there may be a whole new set of patterns that emerge and change workflows such as yours.

Thanks for the feedback!

031-set-pipeline-step/proposal.md
vito: I don't think I have the full picture here as to why the double-tap is necessary but I don't see this proposal forbidding any part of it so long as there's still a way to configure vars from a file. :thinking:

031-set-pipeline-step/proposal.md
hstenzel: I think this is a bit less flexible than the pipeline resource.

Today, we generate the pipelines.yaml and each pipeline itself from a concourse job that scans a GitHub repository. We then create appropriate pipelines (one per included <repo>:<branch>) and let the concourse pipeline resource deploy the whole thing.

How could we deploy pipelines which are unknown at pipeline-writing time with this proposed system as we can with the pipeline resource?

Additionally, this would have strange behavior across branches. If I described my pipeline in a repository, and I create a release branch or a feature branch, then by definition the same pipeline yaml will be present in each branch. This would be wrong and bad (tm) because now I'd have each branch updating the same artifact -- clearly not what should happen!

Could you elaborate a bit more on how such circumstances could be dealt with?

vito: > I think this is a bit less flexible than the pipeline resource.

It is indeed less flexible, at least in its first iteration. This is just the first step along our roadmap. :slightly_smiling_face: This RFC is not intended to replicate the pipeline resource's entire functionality, my goal is just to provide a teeny tiny primitive that provides basic functionality that we can build on.

For example, one thing that this step *won't* do, as opposed to the `concourse-pipeline` resource, is allow you to configure multiple pipelines all at once. Instead, just like all the other steps, it only deals with one thing at a time. To set a bunch of pipelines at once, you would just run multiple `set_pipeline` steps, just as you would run many tasks. To set a *dynamic* set of pipelines, you'll have to wait until the rest of our roadmap is done, namely the `across` step (#29). Until then I would recommend that you either continue to use the `concourse-pipeline` resource for that specific use case. Theoretically you could generate a pipeline with many `set_pipeline` steps, but that may be too many moving parts. :thinking: 

If you haven't yet, I would recommend checking out the [v10 roadmap post](https://blog.concourse-ci.org/core-roadmap-towards-v10/) and accompanying [slides](https://vito.github.io/slides/v10.html), which should paint a fuller picture of how pipelines will be automated, potentially across branches, by composing the `set_pipeline` step with the `across` step (#29) and instanced pipelines (#34). Finally, 'projects' (#32) will tie it all together.

[Here's a slide showing a project that sets pipelines across all branches.](https://vito.github.io/slides/v10.html#8,4)

Maybe it would help if I had a complete example somewhere showing how this would be used in the future. :thinking: I can add something under 'new implications' since that's where it claims to replace the `concourse-pipelines` resource.

> Additionally, this would have strange behavior across branches. If I described my pipeline in a repository, and I create a release branch or a feature branch, then by definition the same pipeline yaml will be present in each branch. This would be wrong and bad (tm) because now I'd have each branch updating the same artifact -- clearly not what should happen!

I'm not sure what you mean here - this sounds like a workflow problem, not a problem with the `set_pipeline` step, since the same challenges affect the `concourse-pipeline` resource. We faced the same challenge and solved it by extracting our pipeline configs into a separate [`ci` repo](https://github.com/concourse/ci). This repo is what we would convert into a 'project' which configures its branch pipeline template across all of our branches.

Does this all help?

hstenzel: Thanks, I have checked out the roadmap post. I'm excited, but admittedly don't yet fully grok the whole thing.

The workflow problem is exactly what I'm trying to think about. Consider a whole bunch of repositories each of which might have several "interesting" branches all producing and consuming artifacts of different kinds.

What should a pipeline correspond to? Right now, we map a github <account>:<repo> to a concourse <team>:<pipeline> and we have homegrown tools so that there is just one logical pipeline described in github that is deployed several times over in different ways to produce the correct results. In this case, the "correct" results mean mainly getting inputs from the correct places and producing artifacts & loading them in the correct place.

The pipelines have to be deployable properly in different teams and work properly across branches and forks. So we have to be very clear about what goes into the repo itself.

Different variables need to be set based on credentials used, accounts used, administration zone used, etc. None of these belong in source control where they can be forked or branched.

I know these questions are bigger than this feature and I'll stay in the loop on the roadmap, but I also want to bring up these larger issues just to be sure that there will be a natural way to work in these more complicated environments with Concourse.

Thanks!

evanchaoli:
@vito Please consider to support pipeline self set. For example, for the `set_pipeline` step, if configure `pipeline-name: self`, where `self` is a reserved word, the it just do self set. With self set, pipeline name and team are known from the current pipeline.

I see you have create issue #4254 to track implementation of `set_pipeline`. Once this step is in place, it will simplify a lot of our use cases. I think `set_pipeline` step could be done on ATC, just call same function as current set-pipeline API, no need to start any container. If that's true and you are ok, I can contribute a PR, as we eager for this feature.

031-set-pipeline-step/proposal.md
```diff
@@ -0,0 +1,105 @@
...
+  * Keyword use has precedent in `version: every`/`version: latest`/`inputs: all`.
+
+  Cons:
+
+  * Supporting both self-updating pipelines and projects could cause confusion
+    and fragmentation; it doesn't seem wise to have two competing approaches to
+    the same goal.
+  * Given that `self` isn't *critical* (it's easy to work around through
+    templating, i.e. `set_pipeline: ((name))`), is it worth the
+    risk/maintenance?
```
git-committed: We have multiple teams and 100s of pipelines coming from multiple git repos. We are addressing `fly sp` workflow for all changes to pipelines ymls with pipeline_resource and/or a custom built resource with fly binaries listening on these pipelines files/folders. 
Even though we do not have strong preference for using either pipeline name or `self` as set_pipeline values( Good that we support both :) ), we are really excited for the set_pipeline plan. We also go through lot of compliance/audit related activities periodically and pipeline config changes being tracked in the same git repo as the pipeline definitions will help a lot.

osis:
Weird error...

```unknown artifact source: '.' in file path './repo/ci/pipeline.yml'```

changing the file path to ```repo/ci/pipeline.yml'``` fixes it.

**Comment by TheDevMinerTV** [comment]
Works so far, but I would like to see the team name to be able to be changed. Right now I have four teams which all have their own auto-setting pipeline, which I would like to merge into one.

**Comment by DanielJonesEB** [comment]
> Please comment on individual lines, not at the top-level.

Err, that guidance aside...

The old pipeline resource accepted globs in its `vars_files` elements. The new step doesn't. I'm not sure I can make a compelling case for accepting more generic input instead of specific input, but it is a bit more hassle if you've got a versioned file being used as a vars file. In our use case we have to use Minio rather than S3, so we have to have a version number in the file name of assets we track.

I thought I'd flag it in case this is an unintentional reduction in functionality.

**Comment by vito** [comment]
@DanielJonesEB Interesting, definitely not intended. Good call! That worked previously because the resource itself performed glob expansion against its local filesystem. For the `set_pipeline` step that will be a bit more challenging, as we don't have access to a local filesystem. Hmmm. 🤔 I'm going to record that as an open question.

@TheDevMinerTV That's a tricky one. Configuring within the team is somewhat easy because we can assume someone set the *configuring* pipeline within that team and therefore they have authority to set other pipelines. But once you go cross-team, that same assumption is no longer valid - only if you're the `main` team, which feels like a pretty limited use case.

I think this touches on a larger goal of full-stack automation, which might involve things like a `set_team` step and perhaps even a `set_project` step (assuming #32) in the future. I think for now it's best that we leave that out of scope of this issue, but it'd be a super interesting discussion to hold elsewhere. Until we hash that out I don't think we can/should have a `set_pipeline` step that can configure in other teams.

**Comment by TheDevMinerTV** [comment]
Sounds fair :smile:

**Comment by xtremerui** [comment]
Coming from https://github.com/concourse/concourse-pipeline-resource/issues/10, is set_pipeline step gonna support destroy or there will be a destroy_pipeline step in the roadmap?

File: 031-set-pipeline-step/proposal.md [hunk 5]
```diff
@@ -0,0 +1,120 @@
...
+
+# New Implications
+
+## Deprecating `concourse-pipeline` resource
+
+Deprecating the `concourse-pipeline` resource should be the primary goal.
+
+Some of the extended functionality of the resource will not be supported in the name of keeping the `set_pipeline` step design simple and easy to reason about.
+
+For example, the step should only ever configure one pipeline at a time - it should not support the `pipelines:` functionality for configuring a bunch at once.
```
xtremerui, line 115: Is it possible to run set_pipeline step multiple times from a list of configuration to mimic the `pipelines:` functionality? I'm afraid the `concourse-pipeline` resource users can't swtich until they find a way to do so, in which case we can't really make the resource deprecated.

vito:
@xtremerui I don't think we'll need destroying with the advent of pipeline archiving, where you can just remove the step from you config.

031-set-pipeline-step/proposal.md
vito: This should probably be done via #29 - it could even be given a static list of pipeline configs to configure. If it's determined at runtime, that list could could be populated via the `load_var` step (#27) or come from a var source.

031-set-pipeline-step/proposal.md
radoslav-tomov: We have a relatively small set of pipelines, so both ways are ok for us. Honestly, I like the `self` build in reference it is clean, straightforward, and easy to deal with.

031-set-pipeline-step/proposal.md
```diff
@@ -0,0 +1,120 @@
...
+# Answered Questions
+
+none
+
+
+# New Implications
+
+## Deprecating `concourse-pipeline` resource
+
+Deprecating the `concourse-pipeline` resource should be the primary goal.
```
radoslav-tomov, line 111: The feature is very cool but has only a fraction of the scope and capabilities of the `concourse-pipeline` resource.  The area where they differ the most is for example the scope, `set_pipeline` is limited to the same team, unlike the `concourse-pipeline` resource, which allows us to create a pipeline in multiple teams and even different Concourse instances. 

We were thinking of the following use case:

Single multitenant Concourse instance (potentially using the external worker setup to limit teams to specific workers for security and performance reasons). With RBAC which limits people's ability to the basic stuff trigger jobs, start/pause pipelines, see the build results, and no hijacking, pruning and etc.

The idea here is to have a secure Concourse instance, that can handle multiple teams. We plan not to have a DB backup and treat the instance as a disposable unit. All content is defined as code and an onboarding pipeline which handles stuff like team creation, pipeline upload and etc. In the end, we would like to a definition on what should be put in Concourse and a sort of GitOps model, which is based on PR approval and follows immediate onboarding (team created, pipelines uploaded and started).

I know this is a weird scenario and maybe it is not a good idea, but never the less will give you a hint on the stupid things people are planning to do :D .

vito: I should clarify that while it's definitely a goal to deprecate the `concourse-pipeline` resource, it's *not* a goal for the `set_pipeline` step itself to be the sole thing to replace it and all of its functionality. It's just a piece of the puzzle, and I think it's important for that piece to stay small in scope to make it easier to recognize different patterns which may need to be addressed in different ways.

What you're planning to do sounds a lot like what I mentioned here: https://github.com/concourse/rfcs/pull/31#issuecomment-625273371 - I don't think it's weird at all, but I think it probably needs to be explored more in the context of https://github.com/concourse/rfcs/pull/32 which is a bit closer to the GitOps motivations.

radoslav-tomov: @vito spot on as always. Thanks for both hints. I'll try to figure out how this direction can potentially influence/impact our plans. In the past, we have been building tons of scripts to automate certain aspects of setting up individual teams environment and I would like to transition all that into a well-defined workflow.

vito: I wonder if we can unify Teams and Projects. 🤔 If we did this, you could theoretically have a full GitOps stack by configuring the "main project" in `concourse web` and having that project configure further projects with a `set_project` step, which in turn configure their own pipelines and access control (...and perhaps other projects?).

With all of this being driven by a build plan, you could do all sorts of interesting things like provisioning environments or worker pools.

I'll keep thinking about this and amend the Projects RFC. Thanks for your feedback!

jchesterpivotal: I think it's problematic to totally unify. Some folks prefer to think about attaching permissions to people (or roles, teams); others to assets (projects, folders, etc).

Under the hood it's fine to have some generic ACL representation, but I suspect most places think one way or the other.

vito: @jchesterpivotal I'm sure the devil is in the details - I'd like to figure out what those details are.

Those 'devilish details' will likely be around things that latch on to Teams for other functionality, like scoping credential access. But at least for credentials we already have a path forward with `var_sources`. Are there any other examples? 🤔

If you take away the name, a Team is really just a way of associating permissions to pipelines. I don't really see it as attaching permissions to people. (Unless I'm misunderstanding or forgetting something?)

Teams and Projects are both mechanisms for namespacing pipelines - to unify them, you could either add a permissions model to Projects or add the whole Projects functionality to Teams... which is to say that Teams on its own doesn't have much meat on its bones.

We constantly hear about folks trying to automate team/full-stack configuration:

* https://github.com/concourse/rfcs/pull/32#pullrequestreview-268542134
* https://github.com/concourse/rfcs/pull/31#issuecomment-625273371
* https://github.com/concourse/rfcs/pull/31#pullrequestreview-423266671 (this thread)
* https://github.com/concourse/concourse/issues/3208#issuecomment-603395182
  * Just about every time we hear people ask for API access or a browser-less logins flow, it's from this use case.

It feels like there's a lot of pressure pushing these concepts together, and it seems wise to try to limit the number of concepts with overlapping behavior anyway.

The Teams mechanic has evolved over time and never really had a strong identity. You used to configure all of a team's auth in `fly set-team` (including things like client IDs/secrets) and "log in to a team," but now you configure auth cluster-wide and log in as a user and just have access to whichever teams allow your user/group. This would pretty naturally extend to Projects, especially now that user tokens don't encode a list of teams (as of v6.1.0).

Folks are wildly inconsistent in how they use Teams - sometimes they're used to enforce boundaries, but sometimes they're literally just used to group pipelines together, with many teams given the same set of permissions. These different usage patterns and overall weak conceptual identity make Teams a difficult feature to design around; for example, the `fly` UX trips people up all the time because it assumes you'll be largely sticking to one team.

So I see this as an opportunity to rally around a replacement that has a lot of user demand and a more intentional design that lends itself well to Concourse-y GitOps patterns.

I'm jotting down notes and seeing a lot of things on the "pro" side, so forgive me if I'm talking past y'all and steamrolling a bit while this idea is still brewing - I'll try to find some time next week to amend the Projects RFC (or maybe fork it) since this discussion should probably move elsewhere. Curious to hear any thoughts on this until then either way. 🙂

aoldershaw: @vito when you say:

> having that project configure further projects with a set_project step, which in turn configure their own pipelines and access control

could you clarify what you mean by "configure their own access control"? Do you envision access control being defined by the project itself using some configuration file(s)?

If so, what do you think about the `set_project` step defining the access control for the project (much like how `fly set-team` currently defines the access control for the team). So, the main project + main project's access control is configured in `concourse web`, and could have a build plan like:

```
plan:
- set_project: A
  ...
  acl:
  - local_user: some_user
    role: owner
  - github:
      org: concourse
    role: viewer
```

and perhaps it inherits the parent's ACL as owner's of the project?

(full disclosure I don't fully know how RBAC is configured so this may not make sense)

This way, projects themselves don't need to change in scope (besides being able to run `set_project` steps) - much like the current proposal where projects exist within a team, they would just exist and have certain access controls associated to them.

aoldershaw: p.s. FWIW as someone who doesn't actually manage a Concourse deployment, let alone a multi-tenant one, I think this is a really cool idea!

vito: @aoldershaw Yeah - I can see either adding it to the `set_pipeline` step or adding the auth config to the project config itself.

Putting it in the project config would be more in line with how teams are structured now, where teams can update their auth config without requiring an operator.

I think the trade-off is that you would then commit your auth config into your own project repo - to some that's probably a feature, but there may be some situations where that's not ideal. (For example, you want the repo to be public but don't want to expose the identities of those who work on it.)

jama22: I think @vito is onto something here. There's Teams as a way of bundling pipelines and then there's Teams as the sole interface for assigning roles / permissions to users. IMO the latter has always felt clunky and a bit bottom up; but we had to implement it this way when we rolled out Users & RBAC to minimize breaking changes. 

I think the way a user experiences Concourse through the web UI makes a lot more sense: you click login, you declare who you are, and Concourse tells you you can see/do. As an administrator, setting permissions is a bit backwards. You need to first identify the bundle of pipelines that you want to operate on, create that bundle through teams, create your permission matrix and then apply that through `set-teams`. For small-to-medium sized teams I think the permissioning is mostly binary: either a user can see/do everything on a team or it can't. For more mature teams with stricter rules, I think its more about exception setting, and that's almost always user-centric e.g. okay only Jane has the ability to `fly intercept`, everyone else is read only.

All of that is to say that permissions and access control on the team/project level probably won't make too much sense in the long term anyways. Whether it makes sense to combine team/projects for all the other reasons of project/team/pipeline automation though, i"ll leave that to folks with other opinons

jchesterpivotal: Whatever comes of it, I would be wary of reinventing any wheels, _especially_ in the permissions space. I prefer ACLs to RBACs due to their flexibility and precision, but RBACs have the nice property that they are easy to learn and reason about (they just suck for fine-grained permissions schemes).

The domain concepts here seem to be: user, thing that can be in a pipeline or is a pipeline, grouping of pipelines, operation on thing, permission linking subject-object-operation. The key being that a team/project unification makes a lot of sense as "grouping", and could even set itself up for arbitrary nesting, but that some additional effort for the permissions model would be needed to talk about operations and store the ACL triples.

radoslav-tomov: > With all of this being driven by a build plan, you could do all sorts of interesting things like provisioning environments or worker pools.

This one will be particularly of interest. We are moving into the direction of having dedicated "external" workers for each team as there is no other way right now to prevent potential security implications or even the noisy neighbor's considerations.

vito: As a follow-up, I talked with @tlwr in https://github.com/concourse/concourse/pull/5729 and we agreed that we can add an experimental `team:` field to expedite the deprecation of the `concourse-pipeline` resource while allowing us an way out so that it can be removed in the future. I've added this under the 'Experiments' section, along with `set_pipeline: self`.

@radoslav-tomov One additional thing to talk about there is also automatically *un*provisioning those environments when a team goes away. Probably a topic for another RFC though. 🙂

031-set-pipeline-step/proposal.md
```diff
@@ -0,0 +1,184 @@
...
+
+Some of the extended functionality of the resource will not be supported in the name of keeping the `set_pipeline` step design simple and easy to reason about.
+
+For example, the step should only ever configure one pipeline at a time - it should not support the `pipelines:` functionality for configuring a bunch at once.
+
+Similarly, the step should not support fully dynamic configuration (`pipelines_file:`).
```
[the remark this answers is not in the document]
vito: So far, `vars` and `var_files` are implemented, `config_file` is the same as `file:`, `unpaused` is just forced as `true`, `team` is experimentally supported (collecting feedback here: https://github.com/concourse/concourse/discussions/5731), and `exposed` has just been proposed (https://github.com/concourse/rfcs/discussions/75).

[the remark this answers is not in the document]
o-orand: Thanks for answering! 
I should have miss something about dynamic configuration of `set_pipeline`. For me, dynamic config was like `file` for task (https://concourse-ci.org/config-basics.html#schema.file-path), or https://github.com/concourse/concourse-pipeline-resource#dynamic for a single pipeline. So, is it possible to generate `set_pipeline` config using previously executed task, for instance to  select with var_files to apply ?

gid0:
Is it planned to implement a get_pipeline step in the future, to do what the concourse-pipeline-resource was doing as "in"?

vito:
@gid0 There are no plans for this at the moment. If you've got a use case for this, feel free to open a new [discussion](https://github.com/concourse/concourse/discussions)!

PR #32, opened 2019-07-16
RFC: Projects

vito:
[Rendered](https://github.com/vito/rfcs/blob/projects/032-projects/proposal.md)

TODO:

* [ ] flesh out cross-pipeline `passed` constraints - it's only mentioned in passing

### DIFF
--- a/032-projects/proposal.md
+++ b/032-projects/proposal.md
@@ -1,0 +1,276 @@
+# Projects RFC
+
+# Summary
+
+A Concourse project is a namespace for resources, tasks, and pipelines. Projects are bootstrapped from a single resource definition and a build plan which executes when new versions of the resource are found. Projects bridge the gap between small-scale and large-scale CI workflows, helping to smoothen Concourse's learning curve.
+
+# Motivation
+
+* Smoothen the learning curve. Right now the jump from `fly execute` to `fly set-pipeline` is pretty steep. Projects can be a user's first introduction to build plans, which transitions nicely into jobs and pipelines.
+* Enable small-scale projects to have a simple Travis/Circle CI single-build workflow, without needing pipelines at all. This is a scenario where Concourse pipelines can be seen as overkill.
+* Enable medium-to-large-scale projects to fully automate their CI stack by using the project's build plan to configure pipelines via the [`set_pipeline` step](https://github.com/concourse/rfcs/pull/31).
+* Enable cross-pipeline `passed` constraints between pipelines within the project. This will allow large pipelines to be broken up into reusable templates while still supporting fan-in downstream.
+* Ensure tasks and pipelines are always in sync by having both of them come from the project instead of passing task configs as a resource throughout the pipeline.
+
+# Proposal
+
+A Concourse project allows you to bootstrap and automate your CI/CD stack by configuring a single resource which contains your project's config:
+
+```sh
+fly set-project \
+  --project booklit \
+  --type git \
+  --source uri=https://github.com/vito/booklit \
+  --source branch=projects-mock \
+  --path ci
+```
+
+The resource must contain a project config as `project.yml` under the configured `--path`.
+
+This path also contains resource, resource type, task, and pipeline definitions for the project. A full project config directory may look something like this:
+
+```sh
+ci/project.yml
+ci/resource-types/github-status.yml
+ci/resources/booklit.yml
+ci/resources/booklit-status.yml
+ci/tasks/test.yml
+ci/tasks/build.yml
+ci/pipelines/booklit.yml
+```
+
+## Anatomy of a `project.yml`
+
+A minimal project config has a `name` and a `plan`:
+
+```yaml
+name: ci
+plan: # ...
+```
+
+The `name` should be short and sweet - similar to a resource name. The project config directory will be available to every build executed within the project, as an artifact named after this value. ([Why only the config directory?](#exposing-only-the-config-directory))
+
+Concourse will continuously check the project resource for new versions and execute the project's `plan` when a new version is found. All pipelines, tasks, resources, and resource types defined in the project will be usable by the project's `plan`.
+
+Within the `plan`, you may choose to just run a task, using projects as a simple pipeline-less CI flow:
+
+```yaml
+name: ci
+plan:
+- get: booklit
+- task: test
+```
+
+...or you may choose to leverage the `set_pipeline` step to automate your pipeline configuration:
+
+```yaml
+name: ci
+plan:
+- set_pipeline: booklit
+```
+
+New project builds can be triggered by `get` steps with `trigger: true`, however `get` steps cannot specify any `passed` constraints:
+
+```yaml
+name: ci
+plan:
+- get: booklit
+- get: golang
+  trigger: true
+- task: test
+  image: golang
+```
+
+If your pipelines use templating, you can automate that just by using a `task`:
+
+```yaml
+name: ci
+plan:
+- task: generate-pipeline
+- set_pipeline: my-pipeline
+  file: generated-pipeline/pipeline.yml
+```
+
+This versatility should help smoothen Concourse's learning curve - there now exists a simpler CI/CD workflow suitable for side-projects and simple workflows that don't really need pipelines at all, and along the way we've enabled arbitrary automation needs for more complex workflows.
+
+## Project-level resources
+
+Projects define resources under the `resources/` directory. These resources can be used by all pipelines within the project.
+
+Resources defined at the project level can be propagated between pipelines using `passed` constraints. This is not possible with pipeline-local resource definitions.
+
+## Pipeline-local resources
+
+Pipelines can still define their own resources in addition to the project scope.
+
+There is some risk for confusion as developers will have two places to look for resource definitions, but the benefits seem to outweigh the risk:
+
+* Pipeline-local resources can be used to prevent polluting the project-wide resource scope with resources that are only ever useful for a particularly specialized pipeline.
+* Pipeline-local resources are natural fit for pipeline templates which use `((vars))` to configure their resources. For example, a branch pipeline template would have a pipeline-local resource definition for its branch.
+
+Concourse will prevent 'resource shadowing' - any pipeline which defines a local resource matching the name of a project-wide resource should error upon configuration.
+
+## Project-level tasks
+
+Projects define tasks under the `tasks/` directory.
+
+Tasks will automatically propagate to every build plan executed within the project via the project's config directory artifact.
+
+The `task` step will be modified to load the task config corresponding to the task name from the project artifact by default.
+
+For example, the following build plan...:
+
+```yaml
+plan:
+- task: foo
+```
+
+...will be interpreted as this:
+
+```yaml
+plan:
+- task: foo
+  file: project-name/tasks/foo.yml
+```
+
+By automatically propagating task configs and giving meaning to the name (`foo`), pipeline configs get a bit cleaner: users no longer have to manually pass around a resource for their CI config, and tasks no longer have to specify `file:`.
+
+Before:
+
+```yaml
+plan:
+- get: ci
+  passed: [unit]
+- get: concourse
+  passed: [unit]
+- in_parallel:
+  - task: fly-linux
+    file: ci/tasks/fly-linux.yml
+  - task: fly-darwin
+    file: ci/tasks/fly-darwin.yml
+  - task: fly-windows
+    file: ci/tasks/fly-windows.yml
+```
+
+After:
+
+```yaml
+plan:
+- get: concourse
+  passed: [unit]
+- in_parallel:
+  - task: fly-linux
+  - task: fly-darwin
+  - task: fly-windows
+```
+
+Additionally, Concourse will keep track of which version of the project was used for the execution of the build and ensure it uses the same version if the build is re-triggered.
+
+## API relationships
+
+Teams may have many projects. Projects are cheap and in many cases will replace `fly set-pipeline`. Users are already configuring many distinct pipelines within a team, so it's natural to expect the same of projects. By existing within a team, projects can have simple names without conflicting with projects elsewhere on the Concourse cluster.
+
+Projects have many pipelines. Projects provide a namespace for pipelines, preventing pipeline names from colliding with other pipelines within the same team. This allows pipeline names to be short and simple.
+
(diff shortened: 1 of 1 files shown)


032-projects/proposal.md
```diff
@@ -0,0 +1,274 @@
...
+## Pipeline-local resources
+
+Pipelines can still define their own resources in addition to the project scope.
+
+There is some risk for confusion as developers will have two places to look for resource definitions, but the benefits seem to outweigh the risk:
+
+* Pipeline-local resources can be used to prevent polluting the project-wide resource scope with resources that are only ever useful for a particularly specialized pipeline.
+* Pipeline-local resources are natural fit for pipeline templates which use `((vars))` to configure their resources. For example, a branch pipeline template would have a pipeline-local resource definition for its branch.
+
+Concourse will prevent 'resource shadowing' - any pipeline which defines a local resource matching the name of a project-wide resource should error upon configuration.
```
aegershman: I don't have any formed opinion on this, but I'm curious if there's reasoning/past examples behind why Concourse will prevent resource shadowing?

032-projects/proposal.md
```diff
@@ -0,0 +1,274 @@
...
+
+Even worse - you could potentially make a mistake and have your project tasks always running against the code from the *project* resource, and not what's actually under test in the pipeline.
+
+So, what's the solution here? Really, it comes down to this: your product code and your CI config are two different things. This becomes clear right from when you had to write an explicit resource for it in order to use `passed` constraints. Applying `passed` constraints to the project itself doesn't really make much sense - the job config had to come from somewhere in the first place to learn about the `passed` constraints. It's a bit of a chicken-egg problem.
+
+Instead, these should be treated as two separate inputs. Your product code is an input alongside your CI code. This also lets you split the CI code out into a separate repo if you want, and use it for every branch of your product, rather than having a copy on each branch and risk having them fall out of sync.
```
aegershman: Unsolicited thought: presumably this allows separating CI from product code && using it for each branch, but also for using the same CI code for _multiple products_ with _multiple branches_ (e.g., using single config repo to handle  CI for like... 13 other git repos, each containing a microservice composing the entire software solution).

032-projects/proposal.md
```diff
@@ -0,0 +1,274 @@
...
+
+A Concourse project allows you to bootstrap and automate your CI/CD stack by configuring a single resource which contains your project's config:
+
+```sh
+fly set-project \
+  --project booklit \
+  --type git \
+  --source uri=https://github.com/vito/booklit \
+  --source branch=projects-mock \
+  --path ci
```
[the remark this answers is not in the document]
vito: "Project version" only refers to the versions returned by the project resource's `check` action, i.e. `ref: abcdef` if you were to use a `git` resource.

I hadn't thought about having Concourse monitor the files under the given directory for change; the `--path` is just used so that Concourse knows where to load the project config from. It will rely on the `check` script and run whenever there's a new version returned by it, regardless of whether the specified path actually changed.

If there's concern over executing the `plan:` unnecessarily, it would make sense to configure the `git` resource with `paths: [ci]` in this case:

```yaml
--source paths=["ci"]
```

Alternatively the `ci` path could be split out into a separate repo like we've done (https://github.com/concourse/ci). In that case you wouldn't specify `--path` at all.

Does that make sense? :slightly_smiling_face:

032-projects/proposal.md
vito: I'm just being cautious. I don't really have any past examples as right now pipelines just have a flat resource namespace.

This is the first time adding two levels of scope. I'm usually not a fan of having to look in multiple places for things that go into my CI. GoCD for example lets you set environment variables at multiple levels, so when I'm trying to find where something is set I have to traverse the whole tree.

In this case there seem to be pretty solid justifications for allowing both levels, so I want to mitigate any potential for confusion I can before we paint ourselves into a corner. To me shadowing resource definitions feels just as risky as shadowing variables in a programming language, and lots of languages emit warnings or errors for it. So until someone provides a really solid use case for shadowing vs. just giving one resource a different name or making all the definitions local to each pipeline instead, I think we should prevent it.

Here's a made up example: say my project uses Go 1.12 and defines a project-wide `golang` resource for it. I only intend to support one version of Go at a time, so the name is just `golang` instead of `golang-1.12` as I don't want to have to update all my pipelines every time I bump the version.

Next I bring in a pipeline that already had its own `golang` resource. This one is also set to version 1.12, but it'd be easy to forget about it. A new version of Go comes out, I bump my project to 1.13, but that pipeline's stays at 1.12 and all the while it just says `golang`. It could take a while to notice that kind of thing.

032-projects/proposal.md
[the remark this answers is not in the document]
aegershman: It does make sense, thanks for helping me understand. To be honest I was originally trying to understand exactly where a `project`'s `version` came from at all, but now it's conceptually clear to me 👍 thanks

032-projects/proposal.md
```diff
@@ -0,0 +1,274 @@
...
+* Enable medium-to-large-scale projects to fully automate their CI stack by using the project's build plan to configure pipelines via the [`set_pipeline` step](https://github.com/concourse/rfcs/pull/31).
+* Enable cross-pipeline `passed` constraints between pipelines within the project. This will allow large pipelines to be broken up into reusable templates while still supporting fan-in downstream.
+* Ensure tasks and pipelines are always in sync by having both of them come from the project instead of passing task configs as a resource throughout the pipeline.
+
+# Proposal
+
+A Concourse project allows you to bootstrap and automate your CI/CD stack by configuring a single resource which contains your project's config:
+
+```sh
+fly set-project \
```
ari-becker: Maybe this is a bit too much scope-creep, but I would propose including some mechanism for bootstrapping Concourse with a list of pre-defined projects - some kind of `initial_config.yaml` that includes teams and projects belonging to those teams, that can be checked in with whatever is used to setup Concourse in the first place.

As somebody who loves Concourse's strongly opinionated commitment to reducing state - [`worker-state`](https://github.com/concourse/concourse/wiki/Anti-Patterns#worker-state) being a recognized anti-pattern - it would be nice to see a nod to trying to reduce `web-state` or `atc-state`. One of the advantages of using the pipeline resource over `fly set-pipeline` was having one single `meta` pipeline, checked into version control, which bootstrapped the rest of the pipelines. Is there any way here to have one single thing of some kind which can bootstrap the rest of the projects on the `atc`?

vito: @pivotal-jwinters, @pivotal-jamie-klassen and I chatted about this as sort of the natural conclusion to this trend. I think it's a great idea but we should get there in the future as a separate RFC. It doesn't replace 'projects' since as you pull on the thread the required role of the user making changes goes from 'team member' to 'team admin' to 'concourse operator'.

So much in the same way that Pipelines were manually configured until Projects were introduced to bootstrap them, I imagine Projects will be manually configured until we introduce something higher up. Eventually we'll end up with all the `web` config coming from a `git` repo, I imagine. :stuck_out_tongue:

ari-becker:
@vito I think that would be a great idea.

On the required role of operator, I'd argue that there's two perspectives:
only requiring team members is great for flexibility, and flexibility is
certainly something that Concourse has excelled at - i.e. `fly execute`. On
the other hand, if the flip side of requiring the cluster operator is that
projects, pipelines, etc. simply *exist*, completely automatically
according to templates, including required verification, notification, etc.
steps, having been setup automatically when new repositories were created,
then, the flexibility just isn't needed at all. I think Concourse can lend
itself very well to the second model as well - usable either way, depending
on the culture and needs of the group running the Concourse instance.

032-projects/proposal.md
ari-becker: @vito I think that would be a great idea.

On the required role of operator, I'd argue that there's two perspectives:
only requiring team members is great for flexibility, and flexibility is
certainly something that Concourse has excelled at - i.e. `fly execute`. On
the other hand, if the flip side of requiring the cluster operator is that
projects, pipelines, etc. simply *exist*, completely automatically
according to templates, including required verification, notification, etc.
steps, having been setup automatically when new repositories were created,
then, the flexibility just isn't needed at all. I think Concourse can lend
itself very well to the second model as well - usable either way, depending
on the culture and needs of the group running the Concourse instance.

vito: Right. What you're describing sounds like a single-tenant use case, where everyone involved in the Concourse installation is aware of how it's all set up and has a workflow that encompasses the whole Concourse cluster. This is true of a lot of (maybe most) Concourse installations.

So far, I'm approaching this from a multi-tenant viewpoint where things are more 'self-service' and users who may not know each other are sharing a Concourse and need to be empowered to configure their own things without having admin capabilities or a central authority merging pull requests in a Concourse cluster automation repo. I think this will be easier to work with if/when we ever want to do a hosted Concourse offering.

Side tangent: back in the day, Concourse was literally configured with a single pipeline. It wasn't even called a pipeline, it was just a config file. I resisted going any further than that for a while, wishing people would just run a different Concourse process for each pipeline, but ultimately that grew into multiple pipelines configured via the API, and from there it grew into multiple teams, etc. etc. etc.

Expanding API flexibility seems to be the trend, and I do sometimes wonder how much simpler some things could be if we enforced one-pipeline-per-Concourse, but I kind of suspect it would just push the complexity into other places. :thinking:

ari-becker: > Side tangent: back in the day, Concourse was literally configured with a single pipeline. It wasn't even called a pipeline, it was just a config file. I resisted going any further than that for a while, wishing people would just run a different Concourse process for each pipeline... 
>
> ... I do sometimes wonder how much simpler some things could be if we enforced one-pipeline-per-Concourse, but I kind of suspect it would just push the complexity into other places. 🤔

Sounds like the Unix philosophy to me - write simple programs that do one thing and do it well, and then compose them to solve more complex problems. The law of conservation of complexity, in context, argues that "pushing the complexity into other places" makes it easier to deal with that complexity by enabling the project to take benefit from the effort that went into the tools in those "other places", instead of re-inventing the wheel.

vito: Yeah, that was the idea. I don't quite remember what was the final straw that pushed us in this direction, but it may have been as simple as wanting to access all of the pipelines in one UI at one URL and with one authentication mechanism.

I think there's a lot of value in the current approach from a user-level simplicity standpoint, and we can still have simple UNIXy principles underlying Concourse as part of its internal architecture, with the existing multi-tenant platform providing a consistent opinionated pattern for Concourse users.

I think now we're into kind of a deep philosophical discussion that probably doesn't have much resolution, so I'll mark this conversation as resolved so it doesn't dominate the review process. :slightly_smiling_face: Fun chat though.

032-projects/proposal.md
```diff
@@ -0,0 +1,274 @@
...
+plan:
+- task: foo
+```
+
+...will be interpreted as this:
+
+```yaml
+plan:
+- task: foo
+  file: project-name/tasks/foo.yml
```
jchesterpivotal, line 131: I'd suggest the taskdir style: `task: foo` would be interpreted by default as `project-name/tasks/foo/task.yml`.

The basic reasons are that tasks can grow large and complex in their own right, so establishing subdirectories in advance saves messy top-level task directories. It's also easier to do things in bulk (eg, use `xargs` to update multiple task YAMLs simultaneously).

wanderanimrod:
@vito , if projects replaced teams, would RBAC and auth move to projects? If so, I think it will make setup and management of RBAC complicated for teams that work on multiple projects.

032-projects/proposal.md
```diff
@@ -0,0 +1,330 @@
+# Summary
+
+A Concourse project is a namespace for resources, tasks, and pipelines. Projects are bootstrapped from a single resource definition and a build plan which executes when new versions of the resource are found. Projects bridge the gap between small-scale and large-scale CI workflows, helping to smoothen Concourse's learning curve.
```
c4lliope, line 3: I usually consider the phrase "project" to be as ambiguous as can be; it gives a vague sense of a user's underlying goals, though offers no indication of what they could be, and in many places and programs is used only to convey a sense that you can draw a clean boundary around a group of resources.

Accordingly - could you consider renaming the "project" idea, using "scope" or "namespace" (as you describe in this line). Perhaps "codebase" could do also, if they are usually to be tied to a git repository like the example suggests.

"Scope", "Namespace", "Space", "Codebase"... - also clearly indicate clean boundaries around your included resources and jobs, but they don't imply as much of an underlying goal as "project" seems to, which leaves room for them to change into a more flexible programming model.

As `Naming` is a hard problem, please use care in choosing a good, clearly-explained name for the ideas that this RFC introduces. As you say, this will likely be the opening concept by which new users learn the workings of Concourse; making a solid choice here could set you apart from similar programs, in terms of how you (and your users) organize mechanical jobs in general.

Nice proposal; much obliged.

032-projects/proposal.md
```diff
@@ -255,6 +255,59 @@ Instead, these should be treated as two separate inputs. Your product code is an
 TODO: finish this thought - wrap it up by suggesting that 
 
 
+# Open Questions
+
+## Can Projects replace Teams?
```
mvdkleijn, line 260: Please don't? 😄 

We use the Teams concept to map onto our product teams. The term "project" is considered a legacy, non-agile term. As such, the first question replacing Teams by Projects would raise in our organization is: "It no longer supports agile organizations?" (I know, I know... still...) We barely convinced people that agile is good and barely got people used to product teams, etc.

I was kinda hoping Projects would become an organizational tool under Teams for housing related Pipelines. I.e. you'd have a Team like "Tasty Treats app" which could have Projects like "Strawberry Icecream", "Bubble Tea" and "Apple Pie" with Pipelines underneath those related to the various tasks and builds that need doing there.

But feel free to ignore my monologue. Just dropping my 2 cents. 😄

c4lliope:
I had a go renaming the ideas in your proposal as "pipeline plan",
each one undergoing a "build phase".
The idea really gains some legs, because the only required keys
inside your `.yaml` are `name:` and `plan:`.

Here is my copy [rendered](https://github.com/c4lliope/rfcs/blob/pipeline-plans/032-projects/proposal.md).
I keep all logical ideas in place and only change the proposal's `project` labels.

fshuva:
Hi, 
I was wondering if there are any plans to add variables in the `set_pipeline` name. If someone wants to use dynamically named pipelines, the only way to do it currently is using a shell script with fly. If we had vars support in the pipeline name, it would be easier to create dynamic pipelines based on passed variables.

dR3b:
hi@all Are you still working on this?

PR #33, opened 2019-07-16
RFC: archiving pipelines

vito:
[Rendered](https://github.com/vito/rfcs/blob/archiving/033-archiving-pipelines/proposal.md)

### DIFF
--- a/033-archiving-pipelines/proposal.md
+++ b/033-archiving-pipelines/proposal.md
@@ -1,0 +1,30 @@
+# Archiving Pipelines RFC
+
+# Summary
+
+This proposal outlines a relatively straightforward 'archiving' operation which can be used to soft-delete a pipeline while preserving its data for later perusal.
+
+# Proposal
+
+Pipelines can be archived by a user through the `fly` CLI:
+
+```sh
+$ fly -t ci archive-pipeline -p pipeline-name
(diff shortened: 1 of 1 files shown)


033-archiving-pipelines/proposal.md
```diff
@@ -0,0 +1,84 @@
...
+# Proposal
+
+Pipelines can be archived by a user through the `fly` CLI:
+
+```sh
+$ fly -t ci archive-pipeline -p pipeline-name
+pipeline 'pipeline-name' archived
+```
+
+Archived pipelines are permanently paused - no resource checking or job scheduling is performed. They should consume no scheduling resources - only database space for the build history. Note that, with the exception of the behaviour of `CreateJobBuild` (outlined below), any behaviour associated with paused pipelines should be inherited by archived pipelines (things like resource checking and garbage collection included).
```
taylorsilva, line 14: Should the logs for build history be retained indefinitely? Should they still be reaped if reaping is configured?

vito: I think it makes sense for them to continue the original build log retention policy. Or at least that feels less surprising than keeping them around forever. Depending on the policy, they *may* stick around forever (i.e. "keep last N builds"), but I don't think archiving should enforce that.

PR #34, opened 2019-07-16
RFC: instanced pipelines

vito:
[Rendered](https://github.com/vito/rfcs/blob/instanced-pipelines/034-instanced-pipelines/proposal.md)

### DIFF
--- a/034-instanced-pipelines/proposal.md
+++ b/034-instanced-pipelines/proposal.md
@@ -33,11 +33,9 @@ fly archive-pipeline -p branch -i branch:feature/foo
 
 ## Automatic archival
 
-When used with the `set_pipeline` step ([RFC #31](https://github.com/concourse/rfcs/pull/31)), instanced pipelines take on additional behavior.
+Instanced pipelines build on the automatic pipeline archiving introduced in [RFC #33](https://github.com/concourse/rfcs/pull/33). Individual pipeline instances that are no longer configured will be automatically archived in the same way that normal pipelines would.
 
-At the end of a build which uses `set_pipeline`, all instances of the named pipelines which were not configured by the build will be automatically archived.
-
-For example, say I have a job which I use to automatically configure my pipelines:
+For example, say I have a job whose build plan configures a pipeline instance for each supported version:
 
 ```yaml
 plan:
@@ -52,7 +50,7 @@ plan:
     version: 5.2
 ```
 
-Let's say I ship a `5.5` version. Assuming my policy is to only support the last 2 versions, I would update the config like so:
+Let's say I ship a `5.5` version, and my policy is to only support the last 2 versions. I would update the config like so:
 
 
 ```yaml
@@ -70,6 +68,7 @@ plan:
 
 When this build runs, the `version: 5.2` instance will be automatically archived.
 
+
 # New Implications
 
-This functionality will be more and more useful as we expand Concourse's vocabulary to support pipeline automation. Spatial resources ([RFC #26](https://github.com/concourse/rfcs/pull/26)), for example, can be used to automatically configure a pipeline for each branch or PR. When the branch or PR goes away, their pipeline instances will be archived automatically.
+This functionality will be more and more useful as we expand Concourse's vocabulary to support pipeline automation. Spatial resources ([RFC #29](https://github.com/concourse/rfcs/pull/29)), for example, can be used to automatically configure a pipeline for each branch or PR. When the branch or PR goes away, their pipeline instances will be archived automatically.
--- a/034-instanced-pipelines/proposal.md
+++ b/034-instanced-pipelines/proposal.md
@@ -1,0 +1,75 @@
+# Summary
+
+Instanced pipelines group together pipelines which share a common template configured with different ((vars)). They provide a simple two-level hierarchy and automatic archiving of instances which are no longer needed.
+
+# Proposal
+
+Pipelines can be configured with 'instance vars' like so:
+
+```yaml
+fly set-pipeline -p branch --instance-var branch=feature/foo
+```
+
+This will configure a `branch` pipeline, with the `((branch))` var set to `"feature/foo"`.
+
+Instance vars are used as part of the pipeline identifier in the UI and API. There can be multiple instances of a pipeline with the same name:
+
+```yaml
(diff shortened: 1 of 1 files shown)


034-instanced-pipelines/proposal.md
```diff
@@ -0,0 +1,75 @@
...
+fly set-pipeline -p branch --instance-var branch=feature/foo
+```
+
+This will configure a `branch` pipeline, with the `((branch))` var set to `"feature/foo"`.
+
+Instance vars are used as part of the pipeline identifier in the UI and API. There can be multiple instances of a pipeline with the same name:
+
+```yaml
+fly set-pipeline -p branch --instance-var branch=feature/foo
+fly set-pipeline -p branch --instance-var branch=feature/bar
```
Zanadar: I think the implicit looping of the configuration here could be supported well in a vars file api too, eg

```yaml
fly set-pipeline -p branch --instance-var-file vars.yml

#! vars.yml
---
branch:
- feature/foo
- feature/bar

File: 034-instanced-pipelines/proposal.md [hunk 2]
```diff
@@ -0,0 +1,75 @@
...
+```
+
+Instanced pipelines sharing the same name will be grouped together in the web UI.
```
Zanadar: This is great! Some sort of hierarchical display would be awesome. https://buildpacks.ci.cf-app.com/ has lots of "instanced" pipelines we create through `erb.yml` craziness, and being able to visually organize them better would be super!

034-instanced-pipelines/proposal.md
```diff
@@ -0,0 +1,74 @@
...
+
+```yaml
+plan:
+- get: ci
+- set_pipeline: release
+  file: ci/pipelines/release.yml
+  instance_vars:
+    version: 5.3
+- set_pipeline: release
+  file: ci/pipelines/release.yml
```
evanchaoli, line 48: How about if I use a different pipeline yml here? Is that allowed? 

- If yes, then this feature is just a way to group pipelines with a single name. In that case, the existing `var` and `var_file` is good enough, no need to introduce `instance_vars`.
- If not, then why let users to input `file` multiple times? The syntax may change to something like:

```yaml
- set_pipeline: release
  file: ci/pipelines/release.yml
  instance_vars:
  - version: 5.3
    foo: bar
  - version: 5.2
    foo: bar
```

vito: The `set_pipeline` step sets all instances of a pipeline at once. They don't technically have to share a common template, that's just how most examples are framed as that seems like the most obvious use case.

Your second example would be accomplished by way of the `across` step outlined in #29:

```yaml
across:
- from: 3.0
  to: 5.0
- from: 4.0
  to: 5.0
as: upgrade_path
do:
- set_pipeline: upgrade-test
   file: ci/pipelines/upgrade-test.yml
  instance_vars:
    upgrade_from: ((upgrade_path.from))
    upgrade_to: ((upgrade_path.to))
```

...though I'm currently leaning towards this syntax instead, since it can be done by just adding the `matrix:` field to any step rather than the full `across:/as:/do:`:

```yaml
set_pipeline: upgrade-test
file: ci/pipelines/upgrade-test.yml
instance_vars:
  upgrade_from: ((.:upgrade_path.from))
  upgrade_to: ((.:upgrade_path.to))
matrix:
  upgrade_path:
  - from: 3.0
    to: 5.0
  - from: 4.0
    to: 5.0
```

In principle, we could make the `file:` field also parameterized through of the matrix if that sort of thing is necessary:

```yaml
set_pipeline: upgrade-test
file: ci/pipelines/((.:upgrade_path.file))
instance_vars:
  upgrade_from: ((.:upgrade_path.from))
  upgrade_to: ((.:upgrade_path.to))
matrix:
  upgrade_path:
  - from: 3.0
    to: 5.0
    file: old-upgrade-test.yml
  - from: 4.0
    to: 5.0
    file: new-upgrade-test.yml
```

So for that case there's probably merit in #5277 after all, though I still don't think we should allow the name to be interpolated as it really overlaps with the intended function of instanced pipeline names; `instance_vars` effectively become part of the identifier, and they don't require you to sanitize whatever you're interpolating since the values aren't part of the pipeline name and thus don't have to conform to #40.

034-instanced-pipelines/proposal.md
```diff
@@ -0,0 +1,74 @@
...
+fly destroy-pipeline -p branch -i branch:feature/foo
+fly pause-pipeline   -p branch -i branch:feature/foo
+fly archive-pipeline -p branch -i branch:feature/foo
+```
+
+(Side note: `:` vs. `=` is a little weird but it's consistent with `fly check-resource` - we use `=` for assignment and `:` for partial filtering.)
+
+## Automatic archival
+
+Instanced pipelines build on the automatic pipeline archiving introduced in [RFC #33](https://github.com/concourse/rfcs/pull/33). Individual pipeline instances that are no longer configured will be automatically archived in the same way that normal pipelines would.
```
evanchaoli, line 36: I'm confused about how to determine "individual pipeline instance no longer configured". My use case is like:

```yaml
- get: policy # this will generate a pipeline yaml on the fly
  trigger: true
- load_var: id
  file: policy/id.txt
- set_pipeline: some-name
  file: policy/pipeline.yaml
  instance_vars:
      id: ((.:id))
```

In this case, `id` is dynamically generated, after build 2, will pipeline set by build 1 be automatically archived?

vito: Yep. This is more a behavior of the `set_pipeline` step in general than with instanced pipelines in particular though.

If the previous pipeline should still exist, that must be expressed by the build continuing to run `set_pipeline` against it.

For example, pull request pipeline automation will continuously re-configure all pipelines for all PRs that are open, rather than setting one-pipeline-per-build:

```yaml
set_pipeline: pr
instance_vars: {pr_number: ((.:pr.number))}
matrix: {pr: concourse-prs}
```

Just guessing from your example, it seems like it should be configuring a matrix of pipelines across the valid policy IDs, similar to above.

evanchaoli: Cool. Thanks for the explanation. Looking forward to this feature.

PR #35, opened 2019-08-09
Proposal for Operator User separation

YoussB:
An idea that arose when working on the super admin story concourse/concourse#4192.


jama22:
In general, I'm not a fan of associating privileges to specific teams. I know we currently have that today with members of the `main` team granting app-wide privileges, and I'm also not OK with that either!

Specifically, I think it creates a lot of surprise and confusion to attach special actions to not just the role that you possess, but also your membership into specific teams. 

Now, the limitation today is that roles are constrained to membership into a team:
User -> Team -> Role 

Maybe its more interesting to introduce new roles that do not need to be attached to the team:
User ->  Role -> Team -> Role

We'd have to design the Roles & Permissions so that the permissions themselves  do not overlap and intersect in surprising ways

PR #37, opened 2019-09-19
RFC: Prototypes

### LINKED ISSUE #7: Reusable tasks

vito:
[Rendered](https://github.com/vito/rfcs/tree/prototypes/037-prototypes/proposal.md)

fixes #7 
supercedes #24


037-prototypes/proposal.md
```diff
@@ -0,0 +1,544 @@
...
+  // Path to a file into which the message handler must write its response.
+  ResponsePath string `json:"response_path"`
+}
+
+// MessageResponse is written to the `response_path` for each object returned
+// by the message. Multiple responses may be written to the same file,
+// concatenated as a JSON stream.
+type MessageResponse struct {
+  // The identifying properties for the object.
+  Object Object `json:"object"`
```
cirocosta: I wonder if there's room for a typed "error" in the `MessageResponse` 🤔 

E.g., if a `git` prototype receives a check message that asks for versions of a `svn` "repository", it could report back that error in the form of a typed "error" that could be gracefully interpreted 🤔 

or, at least a base "error prototype" that can be identified as an error but augmented however one sees fit

037-prototypes/proposal.md
```diff
@@ -0,0 +1,544 @@
...
+  Value string `json:"value"`
+}
+```
+
+
+## Prototype Info
+
+Prior to sending any message, the default command (i.e.
+[`CMD`](https://docs.docker.com/engine/reference/builder/#cmd)) for the image
+will be executed with an `InfoRequest` piped to `stdin`. This request contains
```
cirocosta: I wonder if if we broke from the current flow of piping into `stdin` and writing to `stdout`, things would be simpler when it comes to systems where it's easy to put things on the filesystem (and read results from it later) - 👀 k8s runtime.

maybe this is totally crazy, but, could we even argue that perhaps the process boundary is not the right one? (this is from someone who has been reading about https://github.com/fastly/lucet 😅 where sandboxing/"containerizing" happens within the same process)

037-prototypes/proposal.md
```diff
@@ -0,0 +1,467 @@
...
+specifies the prototype interface version that the prototype conforms to, an
+optional icon to show in the UI, and the messages supported by the given
+object.
+
+### Example **info** request/response
+
+Request sent to `stdin`:
+
+```json
+{ "object": { "uri": "https://github.com/concourse/concourse" } }
```
ari-becker: It's unclear to me where the `uri` field is coming from (specifically for an `InfoRequest`)? If there's any such field to be passed, wouldn't it be something more akin to `ATC_EXTERNAL_URL`? What about the other environment variables that Concourse sets today for `get/put` messages - why are these messages special in that they would receive environment variables that other messages don't receive?

vito: The `object` field is a black box object. In this case it's similar to the `source:` of a `git` resource in a pipeline. It is not the ATC external URL. I'll try to clarify this in the doc. Thanks!

037-prototypes/proposal.md
```diff
@@ -0,0 +1,467 @@
...
+// messages.
+type Object map[string]interface{}
+
+// InfoRequest is the payload written to stdin for the `./info` script.
+type InfoRequest struct {
+  // The properties specifying the object to act on.
+  Object Object `json:"object"`
+
+  // Configuration for handling TLS.
+  TLS TLSConfig `json:"tls,omitempty"`
```
vito: I'm on the fence about including this in `Info`. There may be merit in forbidding network activity in the `Info` call outright. I kind of just threw it in for symmetry.

037-prototypes/proposal.md
vito: :thinking: I agree in principle but I'm having a hard time thinking of cases where the error message would be semantically interpreted and 'handled' in some way. For example I don't really see a practical difference between returning an error message saying "your repo is svn but i'm a git resource" vs. just printing that to `stderr` and exiting nonzero.

Maybe validation errors could be returned as a data structure, but even those could just be printed to `stderr`, and it's kind of hard to specify a data structure here that will capture nuances across all prototypes.

I guess at this point I'm leaning towards under-specifying here and revising the proposal later (it's versioned!) if/when we have more information.

037-prototypes/proposal.md
vito: Scratch that, I forgot gRPC actually is language-agnostic. My concern then would be whether prototype authors would be OK with that overhead vs. simple HTTP.

For HTTP, it could work like this:

* Replace Info with a Start request, which spins up a HTTP server for the given object.
* Messages would be handled by invoking their endpoint, i.e. a `foo` message would be something `PUT /messages/foo`.
  * All endpoints would be technically `PUT` since they perform side-effects and should be idempotent.
* Messages write their response over HTTP via their handler, and no longer have to touch the filesystem. We could leverage HTTP chunked responses here.

I kind of like how the object is represented as a server with its own endpoints. In this setup we don't even have to pass the object twice.

One challenge here though is routing traffic to the server. It's on the container's local network, which is unreachable from the `web` nodes. You would need some kind of proxy between them.

037-prototypes/proposal.md
vito: Regarding `ATC_EXTERNAL_URL` and friends - right now we propagate that to resources in service of notification resources. So I think we can integrate them into a new RFC outlining how prototypes can be used for emitting notifications.

Instead of environment variables, they would probably be injected into the object. This would be done automatically by Concourse whenever a notification is emitted. So notification prototypes could receive an object kind of like this:

```json
{
  "webhook_uri": "https://hooks.slack.com/services/XXXX",
  "notification": {
    "cluster": {
      "url": "https://ci.example.com",
      "name": "my-cluster"
    },
    "build": {
      "id": 1234,
      "status": "succeeded",
    }
  }
}
```

The `notification` field here would be automatically injected into the object by Concourse. The prototype would then read any/all information it needs from it and we can retire the `ATC_*` env vars.

037-prototypes/proposal.md
kcmannem: Would revealing the error as an accessible field allow better control flow in your plan? Such as using the "on_error" hook for a specific type of error from InfoResponse.

kcmannem:
I really like the prototype terminology. It reminds me of smalltalk/objective-c concepts: 
> Every object in Smalltalk, even a lowly integer, has a set of messages, a protocol, that defines the explicit communication to which that object can respond.

It fits really well with the abstractions this introduces.

The challenge I see is communicating these concepts clearly. I know from experience Concourse can have a higher learning curve than other CIs.

I'm excited for this!

037-prototypes/proposal.md
vito: One complication with an HTTP approach: it seems like it would be difficult to re-attach to a message/action that's in-flight or has finished while the `web` node was away.

With writing the result to a filesystem, we can detach from the process and re-attach when we come back, reading the response from the filesystem if the process has already exited.

With HTTP, the client would disconnect, so we'd still need to figure out where to put the response.

At this point I'm leaning towards just making `Info` write its response to the filesystem too. :thinking:

037-prototypes/proposal.md
vito: I suppose this could be useful for verification and for inspecting a remote entity to discover which messages it supports via the prototype. :thinking: 

...but that could also be problematic, as there would be confusion over whether the prototype even implements a message vs. whether the user has permission for the message, and I wouldn't expect all prototype authors to implement 'dynamic message support detection'.

So I'm still leaning towards just removing this and enforcing that `Info` is for local static processing and does not reach out to the network.

037-prototypes/proposal.md
vito: @kcmannem It might, but no one has asked for that feature (to my knowledge). It's hard to design for this kind of thing without having a good grasp of a use case that is also common enough to encode at this level. Personally that amount of control flow/error handling in a pipeline .yml kind of scares me. :sweat_smile:

tremblaysimon:
I don't know if I am out of scope with my comment, but what will be the place of tasks with advent of prototype?

Does it will make sense to continue to do generic task using a pattern like this one exposed here: https://github.com/18F/cg-pipeline-tasks ?

mxplusb:
@tremblaysimon I'm currently on the team that owns the repo you are referencing. Please feel free to ask any questions if you have them.

// cc @bengerman13 @hillaryj

tremblaysimon:
@mxplusb, Thank you for sharing this! I referenced your repo mostly because it's the same pattern we are using at my company (with a private (unfortunately) "concourse-tasks" repo). 

I was wondering if it will continue to make sense to use that type of generic task when prototype will be there. 

In the end, regarding what I understand I think a prototype (like a resource-type today) will be easier to share and use for the consumers.

037-prototypes/proposal.md
```diff
@@ -0,0 +1,473 @@
...
+  - run: build
+    type: oci-image
+    input_mapping: {context: my-image-src}
+    outputs: [image]
+  - put: my-image
+    params: {image: image/image.tar}
+```
+
+This example also makes use of resources, which are also implemented as
+prototypes with special pipeline pipeline semantics and step syntax. These
```
ships: "Special pipeline semantics" really, the invariants I reference below codify a relationship between the high level "pipeline" concept and "resource" prototypes, which together define a category of sorts with desirable structural properties. My open question is, "does affordance for other categories really solve problems for people?" An important part of resources today is that every resource implements the same prototype, and *imply* the abstraction "pipeline" (which is the thing I really want). So while I can imagine an oci image prototype existing, and I can imagine how it "would work" i don't understand what it would "be for" without a complementary high level object that it *implies*.

037-prototypes/proposal.md
```diff
@@ -0,0 +1,473 @@
...
+* It's a pun: 'protocol type' => 'prototype'.
+
+* 'Prototype' has a pretty nifty background in aerospace engineering. This is a
+  completely different meaning, but we never use it that way, so it doesn't
+  seem like there should be much cause for confusion.
+
+
+## Open Questions
+
+* Is this terminology unrelatable?
```
ships: I find it unrelatable but not ambiguous. Given the importance I perceive in either "dependent-types of prototypes" (i.e. any type implementing some set of messages that relate to each other) or "categories" as described above (which include the high level object and its hypothetically invariant properties).

037-prototypes/proposal.md
```diff
@@ -0,0 +1,473 @@
...
+
+
+## Glossary
+
+* **Prototype**: an implementation of the interface defined by this proposal,
+  typically packaged as an OCI container image.
+
+  Examples:
+
+  * a `git` prototype for interacting with commits and branches in a repo.
```
ships: A few comments which together raise the question of how these prototypes should be used. Defining objects might feel nice, but what does it give rise to? A pipeline, which guarantees that a green box at the right means "a pipeline-passing source code and related artifacts were most recently deployed" is extremely valuable to me. While I am already able to violate this by not using well-designed resource types, the prototypes abstraction seems to encourage me to do so. To avoid this, I just need to look clearly at what a pipeline is: its value is *implied* by the structural properties of the resource prototype. I would expect one of a few changes that I would immediately do is re-implement most of the resources I regularly use as children of a general "resource prototype" anyway so that the pipeline is still meaningful to me.

ships: Overall, my review might be addressed in large part by providing an example of another highly useful pairing, comparable to (Resource prototype, pipeline), that I might implement using this new capability.

037-prototypes/proposal.md
```diff
@@ -0,0 +1,473 @@
...
+the prototype itself in response to a prior message.
+
+For example, a `git` prototype may handle a `check` message against a
+repository object to produce commit objects, which then can then handle a `get`
+message to fetch the commit.
+
+An object's supported messages are discovered through an [Info
+request](#prototype-info). For example, the `git` prototype may support `get`
+against a commit object, but not against a repository object. The Info request
+is also a good place to perform validation.
```
ships: This section gives me another way to put my points in other comments, which is, I would rather "assume" things (perhaps with compiler time errors now) about my resources when I use a pipeline. And so you could say that all I want is to introduce a comparable generalization of pipelines, and say that pipelines still only work with the resource prototype-type.

037-prototypes/proposal.md
```diff
@@ -0,0 +1,473 @@
...
+
+* The root of the name comes from [Prototype-based
+  programming](https://en.wikipedia.org/wiki/Prototype-based_programming), a
+  flavor of object-oriented programming, which the concepts introduced in this
+  proposal mirror somewhat closely. There are also some hints of actor-based
+  programming - actors handling messages and emitting more actors - but that
+  felt like more of a stretch.
+
+  This is also an opportunity to frame Concourse as "object-oriented CI/CD," if
+  we want to.
```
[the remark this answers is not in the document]
vito: It's true that Resources have helped reason about Concourse all as one system, but just as there is more to Haskell than just the IO monad (or even Monads in general), there is more to what people are doing with Concourse than what Resources are meant to be used for (see [What makes a resource?](https://blog.concourse-ci.org/reinventing-resource-types/)).

With the advent of this proposal I expect dashboards and pipeline UIs to become more precise. Today we overload Resources beyond essential pipeline metaphors like artifacts and dependencies with other meanings like notifications, deployments, sets of values (e.g. the set of all branches/pull requests). If you cross your eyes and hold your breath you can get these concepts to "almost" be a resource - until they aren't, which is where we've been stuck with a local maxima with user pain ranging from mild (UI cluttered with notifications) to acute (PRs are either hard, through pipeline automation, or easy but with pitfalls).

Fitting everything into one shape worsens the signal-to-noise ratio and means I have to *think* about each object on my screen to understand what it actually is, as they otherwise are shown the same - is this output actually an artifact being delivered, or is it just another resource being used to send a Slack message? Do Slack messages really justify taking up a good chunk of my pipeline UI?

What we're discussing here is a protocol that can support more than just the 'resources' idea - it should not diminish it, it should strengthen what remains, and allow these other concepts be framed properly.

Does that help? 🤔

037-prototypes/proposal.md
vito: > My open question is, "does affordance for other categories really solve problems for people?"

Yes. These problems are cited as motivations for this PR and other PRs which describe other use cases for prototypes:

* https://github.com/vito/rfcs/blob/spatial-resources/029-across-step/proposal.md#motivation
* https://github.com/vito/rfcs/blob/notification-resources/028-notification-resources/proposal.md#motivation

The magnitude of these proposals really cannot be overstated. It is the core focus of the v10 roadmap which is based on years of user feedback and complaints over not properly supporting pull requests and notifications.

Beyond these two concrete workflows, the need for a better way to distribute arbitrary functionality has been brought up since the very early days of Concourse, and is typically cited as concourse/rfcs#7.

There is also a path forward to finally solving https://github.com/concourse/concourse/issues/3023 through the use of prototypes for [https://github.com/vito/rfcs/blob/var-sources/039-var-sources/proposal.md), which would allow credentials to be fetched by using a prototype on a worker with IAM roles configured. This has also been a user need for years.

Further, prototypes can be used to represent things which should trigger builds, but are not necessarily *versioned artifacts* (which imply a central, globally verifiable artifact history), thereby solving the "thundering herd" problem of having many jobs downstream of a 10 minute interval, using the [`get_var` step](https://github.com/vito/rfcs/blob/trigger-resources/027-var-steps/proposal.md).

037-prototypes/proposal.md
vito: I think there may be some confusion underlying a few of these comments; this is an underlying protocol, and not a replacement for resources. None of this should change. Resource types would simply be reimplemented with the same protocol, and pipelines - and your interpreting of their meaning - would stay the same. The only difference is that things which *should not* be represented as resources no longer *have* to be.

vito:
@tremblaysimon Sorry, didn't notice your question there! It is indeed a goal that Prototypes fill the need for generic re-usable tasks. The [Re-inventing resource types blog post](https://blog.concourse-ci.org/reinventing-resource-types/) demonstrates how. 🙂 

The long-term trend I want to see play out is away from tasks and towards prototypes. I think tasks will have their place as the first step: I just want to write a script and run it in an image. Lots of projects probably won't have a need to go past this step. Frankly a lot of CI scripting is ad-hoc.

As patterns emerge, prototypes for common tasks can start to take shape. For example, a `ruby` prototype may emerge which provides basic testing facilities, perhaps even gem building. A `go` prototype could be made for running tests, building binaries, and maybe even monitoring dependencies for new versions. The goal is to gradually replace "glue code" with testable, reusable, shareable prototype implementations.

037-prototypes/proposal.md
ships: First of all, thanks as always for your extremely well-considered and patient thoughts, feedback, and questions. It is such a pleasure and charm to work with you and this project :) 

and in case it does not come across- i am very excited for this update.

Based on further clarifying conversations, i think i had an Aha! moment, around how many of these prototypes we might expect to be in circulation. Per some of your comments above, you might imagine the "resource" concept being separated (rather than conflated, as it is now) with the "notification" concept , by way of each being its own prototype. Off-hand, i would be surprised if there were more than, say, 10 generally useful prototypes at this level of abstraction. I also have difficulty seeing how "git" or "oci image" is sibling to, rather than child of, "resource", when thinking in this way.

The best explanation that comes to mind is that `git prototype` is a toy example and so its function is less clear to me than the real example of `notification/notifier prototype`. And so my confusion would just have arisen around these different examples, and my suggestion i suppose would be to focus on examples like `notification`. If not, could you maybe refer me to a usage of the `git` prototype that would be useful (currently, the example in your protocol doc is more of the how than the why)?

vito: No worries. Thanks for engaging on this - this is a pretty intricate problem space so it's super helpful for finding the parts that aren't totally clear. 🙂

I think what you're suggesting is that these sets of *interfaces* would likely not be many, and I agree. I think the number of interfaces will tend to correspond to the number of common automation workflows and patterns that exist in the software industry. This is probably a smallish number like you said, though I'm still finding [new instances of them](https://github.com/concourse/concourse/discussions/4949#discussioncomment-5681) now and then.

The number of concrete implementations of these interfaces will correspond to the number of distinct tools and services available, which is a much larger number, but still bounded. (There will likely be competing implementations as well.)

However keep in mind this is also to support packaging and running arbitrary code which is not really expected to conform to a particular interface. This makes the set of implementations effectively unbounded.

> I also have difficulty seeing how "git" or "oci image" is sibling to, rather than child of, "resource", when thinking in this way.

Hm, I wouldn't say the `git` prototype is a sibling of a resource, I'd say the `git` prototype *contains* a resource implementation - perhaps even multiple, as we've found that folks want to version based on commit but also based on tags. The `git` prototype can implement objects which support the verbs which make up a resource (`check`, `get`, `put`), but it might also support other verbs which can be done to other types of objects.

Maybe a better example would be a `github-pr` prototype, as it would likely implement three interfaces: a var source, a resource, and a notifier. The end goal is to support enumerating open PRs in a repo (via a var source), discovering and fetching the commits within a PR (via a resource), and updating the status of the PR on GitHub when the build completes (via a notifier).

I'll try to demonstrate that here and maybe work it into the RFC if it goes over well. I'm gonna make up a notation: `some-type{some:object}.some-message` means send the `some-message` message to the `{some:object}` object using the `some-type` prototype.

Note that objects are literally just JSON objects - it's up to the prototype to figure out what the object is describing and what messages the object supports based on the structure of the object itself.

Given a repo object, which only has the `repo` field, it could know how to `list` the set of open PRs against a repo:

```
github-pr{repo:concourse/concourse}.list
=> {pr:123} {pr:124} ...
```

With #29, this response would be used to configure a pipeline for each pull request, using `pr: 123` as a var.

Each pipeline would configure a resource like so, configuring itself as a notifier (this is where you'd otherwise configure a Slack notifier or whatever):

```yaml
resources:
- name: pr
  type: github-pr
  notifiers: [self]
  source:
    repo: concourse/concourse
    pr: 123
```

(Sidebar: an alternative here could be to use the `git` resource prototype to track the PR branch, so as to not duplicate code between `github-pr` and `git`, while defining a separate `github-pr` notifier for the `git` resource. But for the sake of this demonstration let's say it's all baked into one prototype.)

Concourse would send `check` to this object to find commits:

```
github-pr{repo:concourse/concourse,pr:123}.check
=> {ref:commit-1} {ref:commit-2}
```

...And when a build runs, Concourse would `get` a specific commit object with the `get` step, which would run...:

```
github-pr{repo:concourse/concourse,pr:123,ref:commit-1}.get
=> {ref:commit-1} (and bits are fetched)
```

...and when the build completes, the commit object would be extended with build status information, and the notifier would be run, updating the PR status on GitHub:

```
github-pr{repo:concourse/concourse,pr:123,ref:commit-1,build:{id:123,status:"succeeded"}}.notify
=> (no response needed)
```

So in total, this single `github-pr` prototype implements a set of objects which together support three interfaces:

1. The repo object supports the `list` message, and so it could be used with #29 to fan-out across all PRs and set one pipeline for each.
1. The PR object supports `check` and returns commit objects which support `get`, allowing it to be used as a resource in the pipeline.
1. The commit status object supports the `notify` message, tying it in to #28 (which still needs to be fleshed out).

037-prototypes/proposal.md
```diff
@@ -0,0 +1,473 @@
...
+  while we weren't attached:
+  [concourse/concourse#1580](https://github.com/concourse/concourse/issues/1580)
+
+* Support for showing icons for resources in the web UI:
+  [concourse/concourse#788](https://github.com/concourse/concourse/issues/788),
+  [concourse/concourse#3220](https://github.com/concourse/concourse/pull/3220),
+  [concourse/concourse#3581](https://github.com/concourse/concourse/pull/3581)
+
+* Standardize TLS configuration so every resource doesn't implement their own
+  way: [concourse/rfcs#9](https://github.com/concourse/rfcs/issues/9)
```
mxplusb, line 93: Some discussion of this in https://github.com/concourse/concourse/issues/5190, for reference.

PR #38, opened 2019-09-19
RFC: Resource Prototypes

vito:
[Rendered](https://github.com/vito/rfcs/tree/resource-prototypes/038-resource-prototypes/proposal.md)

supercedes #26

### DIFF
--- a/038-resource-prototypes/proposal.md
+++ b/038-resource-prototypes/proposal.md
@@ -1,0 +1,173 @@
+# Resource Prototypes
+
+A *resource prototype* is an interface implemented by a prototype which allows
+it to be used as a *resource* in a pipeline.
+
+
+## Motivation
+
+* Support for creating multiple versions from `put`: [concourse/concourse#2660](https://github.com/concourse/concourse/issues/2660)
+
+* Support for deleting versions: [concourse/concourse#362](https://github.com/concourse/concourse/issues/362), [concourse/concourse#524](https://github.com/concourse/concourse/issues/524)
+
+* Make the `get` after `put` opt-in: [concourse/concourse#3299](https://github.com/concourse/concourse/issues/3299), [concourse/registry-image-resource#16](https://github.com/concourse/registry-image-resource/issues/16)
+
+
+## Proposal
+
+A resource prototype implements the following messages:
+
+* `check`
+* `get`
+* `put` (optional)
+* `delete` (optional)
+
+Their behavior is described as follows.
+
+### `check`: discover new versions of a resource
+
+Initially, the `check` handler will be invoked with an object containing only
+the fields configured by the resource's `source:` in the pipeline. A
+`MessageResponse` must be emitted for *all* versions available in the source,
+in chronological order.
+
+Subsequent `check` calls will be run with a clone of the configuration object
+with the last emitted version object merged in. The `check` handler must emit
+the specified version if it still exists, followed by all versions detected
+after it in chronological order.
+
+If the specified version is no longer present, the `check` handler must instead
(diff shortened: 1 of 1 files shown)


038-resource-prototypes/proposal.md
```diff
@@ -0,0 +1,173 @@
...
+
+
+## Proposal
+
+A resource prototype implements the following messages:
+
+* `check`
+* `get`
+* `put` (optional)
+* `delete` (optional)
```
vlfig, line 23: I fully appreciate I'm rather late to this party and the ship might be ready to sail, but there's one message for resource types to implement that I can see being useful paired with "causality": `diff` (optional, of course).

What's being termed "causality" brings the DAG of resource revisions and the jobs in between them

    (r1)-[input]->(job)-[output]->(r2)

to the presentation layer. This is nice. This allows us to determine all the "things" that went _into_ a given revision of a resource, or vice-versa, all the things a given versioned resource has gone into.

But this _composition_ of being an ingredient to jobs that, perhaps transitively, output a "thing of interest" is rather opaque. Yes, we will have at a glance the exact versions of the libraries and applications and Dockerfiles and helm charts and what-not that made our deployable what it is, but more often than not what the user really wants is the change. The delta. The difference. But diffing our end artifacts will be of no use. For all but the simplest of delivery DAGs we'd only get gibberish. Any meaningful drill down into those differences would require particular knowledge of the structure and semantics of each of the components that went _into_ that artifact.

That knowledge lives in resources. A git resource knows how to interpret and display the differences between two revisions of a git repository. An oci-image resource would know how to diff docker images. Different resources would have a particular interpretation of what it means to be different, because they have the knowledge to _normalize_ before diffing, to hide irrelevant changes or to capture relevant ones from elsewhere if need be.

The causality DAG gives us all the upstream (resp. downstream) resource revisions that went into (resp. used) a given resource revision. If we select two such revisions we have a DAG (or a list if we prefer it simpler) of pairs of revisions for each resource. If each resource could describe the difference between any two of its valid revisions we now have the most meaning-preserving description of the upstream differences "causing" a revision to exist, or the downstream "consequences" of a revision.

The obvious use case is of determining what's about to be deployed, by comparing in this manner the currently deployed artifact and the soon-to-be. It's a different ball game to be able to see immediately in the UI the changes to each component in the way they're most faithful. Imagine seeing in one go the version bump of a library along with improved GC stats, along with a different runtime parameters in the Dockerfile and lower memory requests in the helm chart. Those 'along' quickly become 'because'. Now that's causality.

---

Of course I don't know exactly what the message will look like and what level of structure would be appropriate for the response or the kind of parameterization that would make sense (verbosity level, maybe?). Regrettably I haven't added any pictures to help words make the point across, but if this resonates and there's interest I'm willing to give it a good go and see where that leads us.

vito: That sounds really cool! I wonder if part of the response could be - at least in cases where it's possible - a link to an external service's view of the delta. For example it would be pretty simple to return a link to a commit range diff on GitHub. I'm not really sure what would be possible in other scenarios. 🤔

It does sound like a topic for another RFC, but it's nice to know that the more open-ended nature of the Prototypes protocol could allow for these sorts of things to be implemented and supported in some way without having to worry about whether it has to apply to every single resource type. Thanks for sharing!

vlfig: Thanks!

You make a very good point. External tools might be a good way to present that information. Another question is whether it'd make sense to provide some sort of rendering hint in the diff metadata but maybe that's too strong a coupling with the UI.

I'll create another RFC based roughly on the current one, just for the diff aspect and we'll go from there, then.

PR #39, opened 2019-11-13
RFC: var sources

vito:
[Rendered](https://github.com/vito/rfcs/tree/var-sources/039-var-sources/proposal.md)

supercedes #21


039-var-sources/proposal.md
```diff
@@ -0,0 +1,187 @@
...
+A var source's `type` specifies one of the supported credential managers, e.g.
+`vault`, `credhub`, `kubernetes`. is responsible for interpreting
+`config`.
+
+A var source's `config` is a "black box" to Concourse and is passed verbatim to
+the credential manager. This configuration should include any credentials
+necessary for authenticating with the credential manager.
+
+A var source's `config` may use `((vars))` to obtain its own credentials,
+either by using a different var source or by using the system-level credential
```
evanchaoli: I think a `var_source`'s `config` should only get var from the system credential manager.

039-var-sources/proposal.md
```diff
@@ -0,0 +1,187 @@
...
+Either way, the config will be encrypted when stored in the database.
+
+## `((var))` syntax
+
+The `((var))` syntax was introduced a long while back and was never formally
+specified or documented. This RFC proposes a change to it so now's a good time
+to lay it all out.
+
+The full `((var))` syntax will be
+**`((VAR_SOURCE_NAME:SECRET_PATH.SECRET_FIELD))`**.
```
evanchaoli: Do we want to support multiple var source names? Like: `((vault1,vault2:password))`. This might add some flexibility.

And think we may still support `((secret))` to iterate all var sources.

039-var-sources/proposal.md
```diff
@@ -0,0 +1,187 @@
...
+convenience. This RFC makes no judgment call on this because the utility of
+this will vary between credential managers.
+
+
+# Open Questions
+
+* How should pipeline-level credential managers differ in behavior from
+  system-level credential managers, if at all?
+
+* Should we allow multiple var sources to be configured at the system-level?
```
evanchaoli: I would say no. Because:
1. If the second var source is also owned by the concourse ops team, then not need the second at all.
2. If the second var source is owned by team/pipeline owners, then they will have give the vault access token to concourse ops team in order to add the vault to concourse cluster, which brings a security concern, token might be leaked by concourse ops team.

With pipeline var source, a pipeline owner only needs to add his own vault's token to the system vault. This way, concourse ops doesn't know anything about users' vaults, which eliminate the security concern. This model will reduce OPs cost of concourse clusters.

039-var-sources/proposal.md
vito: Personally I don't see a need for specifying multiple var source names. Seems like the user should know where their secrets are coming from if they know enough to suggest a couple of specific sources. :thinking:

Querying multiple var sources could result in surprises. For example, if their pipeline relied on a credential `foo` from `vault2` but then a credential under the same name was added to `vault1`.

039-var-sources/proposal.md
vito: I don't feel *too* strongly about this, but I decided to allow it so that it's consistent with how `resource_types` works - resource types can use *other* resource types to fetch their image, so it seemed only fair to allow the same of var sources. Do you think this would be particularly challenging to support?

evanchaoli: I feel challenging of implementation. It should be done in `db.pipeline.Varaibles()`, a small change will support that. My concern is that would lead some potential complexity, for example, cycle reference:

```yml
var_sources:
- name: vs1
  type: vault
  config: 
      token: ((vs2:mytoken))  
- name: vs2
  type: vault
  config: 
      token: ((vs1:mytoken))  
```

If we support get var_source's var from other var_source, then we will have to detect such cycle reference problem, maybe other issues also. But `resource_type` doesn't have such problem.

So I think, at least, we start with not supporting that. If users raise such request in future, we may consider to add such support.

039-var-sources/proposal.md
evanchaoli: Make sense.

039-var-sources/proposal.md
vito: `resource_types` actually does have the same problem:

```yaml
resource_types:
- name: foo
  type: bar
  source: # ...

- name: bar
  type: foo
  source: # ...
```

The implementation is pretty simple: if resource type `bar` is being used, then `bar` is ignored when recursively resolving its type `foo`. This trick also allows for e.g. the `registry-image` base resource type to be overridden, using the base resource type to fetch its replacement image:

```yaml
resource_types:
- name: registry-image # same
  type: registry-image # same
  source: # replacement
```

I think the same could probably be done here; instead of going into an infinite loop, an `UnknownVarSourceError` could be returned, and we could print a nice error if the 'unknown source' name is the same as the one we're currently resolving.

In any case, I'm OK with waiting for users to ask for it, since it'd be easy enough to implement later. It just felt like something worth specifying one way or the other, and I can't think of any drawbacks I figured I'd just allow it.

How about I move this under 'Open Questions' and we can collect more feedback on it?

evanchaoli: > How about I move this under 'Open Questions' and we can collect more feedback on it?

That's a good idea.

vito: Moved it to 'open questions'. I have to say as I wrote it out it kind of feels sensible to allow it, primarily for the 'bootstrapping' use case, but we can talk it out there.

039-var-sources/proposal.md
vito: What I'm trying to figure out is if there is clear and consistent behavior that we can figure out in order to support multiple levels of `var_sources`, because that will be something we have to figure out once projects (#32) is here.

Having only a single system-level credential manager kind of feels like an artificial limitation if we're going to support multiple at the pipeline-level, so I'd like to have solid reasoning as to why.

I could imagine, for example, a Concourse installation which has a Vault credential manager as well as an EC2 IAM role-based STS credential manager (concourse/concourse#3023) for access to AWS resources. As this depends on the EC2 configuration of the `web` node, it makes sense for operators to configure it, not pipeline authors. For the same reasons outlined in this proposal, though, it seems like it would be bad to require operators to choose between Vault and STS as the latter is very specialized.

vito: Sounds like this turned out to be useful: https://github.com/concourse/concourse/pull/4777#issuecomment-559631169

I'll update the RFC soon!

vito: Oops nevermind this was a completely different question.

vito: Now that I'm updating the docs for `var_sources`, I'm finding it very difficult to consolidate the docs for cluster-wide credential managers, which are configured through env vars and flags, and pipeline-local `var_sources` which are configured through YAML.

If it's hard to document, it'll probably be hard to learn and use. I think it would be a lot easier to be able to document `var_sources` exclusively and then just say they can be configured cluster-wide or local to a pipeline. This will also force the issue of having to reason about multi-tiered `var_sources`, which I'm anticipating to be something we need to address in the future anyway with Projects.

So at this point I'm leaning towards unifying the cluster-wide and pipeline-local credential manager config as a single concept, `var_sources`, and defining the scoping semantics such that the local var source names 'shadow' the outer ones. We kinda planned for this already anyway in #4777.

jpluscplusm:
Hey @vito 👋 The docs at https://concourse-ci.org/vars.html#var-sources currently say "var_sources [...] is considered an experimental feature until its associated RFC is resolved" with the RFC link being to this, merged, PR. JOOI is the feature still experimental, should the link point elsewhere, or ... something else?

vito:
@jpluscplusm Still have work in flight on this - we're working on a feature so that var sources can be bring-your-own just like resource types so we don't have to bake them into Concourse core code. Once that lands we plan to phase out the built on ones and there may be changes to how they're configured, so it's still experimental for now. More info in concourse/concourse#5229

xtremerui:
the doc is updated to point to the open issue instead of this closed RFC.

jpluscplusm:
@xtremerui Just FYI, I'm not seeing that update online, and I can't see a commit in the docs repo that would suggest a website rebuild is pending. However I'm not using the project actively any more, so I'll stop commenting now.

weaversam8:
confirming that the docs still points to this PR

philocalyst:
Why do the docs refer to this as expiremental when it's merged?

taylorsilva:
@philocalyst it's still marked as experimental because it's implementation is not its final form. That said, I think we should point the docs to https://github.com/concourse/concourse/issues/5229 instead.

philocalyst:
Ah! Missed that.

▰▰▰▰▰
Miles Wirht 🙃

PR #40, opened 2019-11-13
RFC: valid identifiers

vito:
[Rendered](https://github.com/vito/rfcs/blob/identifiers/040-valid-identifiers/proposal.md)

Prior discussion: https://github.com/concourse/concourse/issues/3985

### DIFF
--- a/040-valid-identifiers/proposal.md
+++ b/040-valid-identifiers/proposal.md
@@ -1,0 +1,72 @@
+# Summary
+
+Proposes a fairly limited set of valid characters for use in Concourse identifiers:
+
+* Team names
+* Pipeline names
+* Job names
+* Step names
+* [Var source][var-sources-rfc] names
+
+# Motivation
+
+Concourse does not currently perform much validation for identifiers that can
+be set through the API or in pipeline configs. As a result, users are able to
+name pipelines and jobs with all sorts of special characters.
+
+Allowing arbitrary symbols makes it difficult for Concourse to support semantic
+notation in the CLI and elsewhere in the Concourse UX. For example, the `fly`
+CLI uses `PIPELINE/JOB` syntax, but this becomes ambiguous if your pipeline or
+job names have a `/` present in their name. For example, `foo/bar/baz` could
+either be (`foo/bar`, `baz`) or (`foo`, `bar/baz`).
+
+Allowing whitespace, capitalization, and mixed use of `_` and `-` also results
+in inconsistent naming conventions between Concourse users. A 'deploy to prod'
+job may be called any of the following:
+
+* `deploy to prod`
+* `deploy-to-prod`
+* `deploy_to_prod`
+* `Deploy to prod`
+* `Deploy to Prod`
+
+This variance is largely cosmetic and only makes it difficult for a new
+Concourse user on an existing team to predict the name of a given job. It also
+results in an inconsistent UX when using Concourse.
+
+# Proposal
+
+This proposal is to dramatically reduce the allowed character set for Concourse
+identifiers, in the spirit of simplicity and consistency across Concourse
+usage:
+
+* Any lowercase Unicode letter.
+* Any Unicode number.
+* `-`
+
+With Go's [`re2`](https://github.com/google/re2/wiki/Syntax) syntax this would
+be represented by the following regular expression:
+
+```re
+[\p{Ll}\p{Nd}\-]+
+```
+
+## Migrating existing data
+
+TODO
+
+# Open Questions
+
+* Do we want to support a 'display name'?
+
+
+# Answered Questions
+
+* n/a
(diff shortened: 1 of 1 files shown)


040-valid-identifiers/proposal.md
```diff
@@ -0,0 +1,84 @@
...
+Notably, `_` is forbidden. This is to further enforce consistency in word
+separators.
+
+## Migrating existing data
+
+TODO
+
+# Open Questions
+
+* Do we want to support a 'display name'?
```
ari-becker: For my personal circumstance, I don't need this; however, supporting a display name is probably a requirement to provide an upgrade path for a lot of users. People get attached to their personal preferences and you risk admins facing internal political wrath in their organizations over minor quibbles like this.

040-valid-identifiers/proposal.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Proposal
+
+This proposal is to dramatically reduce the allowed character set for Concourse
+identifiers, in the spirit of simplicity and consistency across Concourse
+usage.
+
+The following characters are to be allowed:
+
+* Lowercase Unicode letters.
```
vito: Hmm, my intent in allowing any lowercase Unicode letter over just `a-z` is to support our multilingual friends, however I just realized that it's hard to enforce "lowercase" when some languages don't have casing at all.

e.g.:

```ruby
irb(main):001:0> /\p{Ll}/ =~ "ひらがな"
=> nil
irb(main):002:0> /\p{L}/ =~ "ひらがな"
=> 0
irb(main):003:0>
```

Of course, we could match all these specific scripts by identifying their Unicode block:

```ruby
irb(main):003:0> /\p{Hiragana}/ =~ "ひらがな"
=> 0
```

...but that seems pretty high-maintenance.

I see two paths forward here:

* Enforce `a-z`. Do people feel strongly about this? I don't think I've ever seen an example of e.g. a Japanese job name.
* Allow uppercase letters and permit this level of inconsistency.
  * If we explicitly allow this we should be sure to not allow e.g. `foO` and `Foo` to both exist. I'm not sure what the situation is now.

vito: Resolving this by allowing all cases and just converting to lowercase.

topherbullock:
This would break Christmas
https://twitter.com/topherbullock/status/950411039954399232

You're a mean one, Mister @vito

nebhale:
I’m pretty wary of this change.  First, I am sympathetic to the technical challenges that arbitrary naming has to parsing and other programmatic processing of the identifiers; it’s a bear and having an unambiguous syntax is not only useful, but preferable.

However, the strict limitation of only Unicode characters plus a pair of word separators (side note, it’s nearly impossible to find a concise listing of what counts as a Unicode “letter”) is a massive regression in _visual_ differentiability. Take the following examples from my own pipelines which contain a great many, very similar resources often times based on different artifacts from the same root project.

```
dependency-github.com-paketo-buildpacks-azure-application-insights
module-github.com-paketo-buildpacks-azure-application-insights
package-github.com-paketo-buildpacks-azure-application-insights
```

Compare those names to an equivalent with an idealized set of names

```
dependency:github.com/paketo-buildpacks/azure-application-insights
module:github.com/paketo-buildpacks/azure-application-insights
package:github.com/paketo-buildpacks/azure-application-insights
```

Or even a slightly sanitized version I’ve used (because it’s not just that `/` is a problem, it’s totally broken in parts of `fly`)

```
dependency:github.com|paketo-buildpacks|azure-application-insights
module:github.com|paketo-buildpacks|azure-application-insights
package:github.com|paketo-buildpacks|azure-application-insights
```

Both of the latter two options are more understandable both in what “type” (dependency, module, package) they are as well as what they refer to (https://github.com/paketo-buildpacks/azure-application-insights)

A display name (now removed from the RFC) is one option here, but I think it’s likely _nearly_ redundant for many identifiers where the majority of characters would be identical, but certain separator characters would be subbed back in for dashes; not a very satisfying way of writing pipeline definitions.

Instead I’d propose that the character set should be loosened to include the common separator characters that are _not_ ambiguous and problematic in the way that `/` is.  There is some concern about the need to quote in a terminal, but that’s a common practice that is familiar to all terminal users and doesn’t feel like a concern that a primarily server-side and web-interfaced application should be ranking strongly.

vito:
@nebhale Good point! I don't think super-long-pipeline-names-with-embedded-values is a pattern we really *want* though - it feels more like a workaround in absence of proper grouping and labeling, which is what instanced pipelines (#34) intends to achieve:

```sh
$ fly get-pipeline -p dependency -i repo:github.com/paketo-buildpacks/azure-application-insights
```

I'll update the RFC to mention this. 🙂

PR #41, opened 2019-11-20
RFC: OPA integration proposal.

evanchaoli:

### DIFF
--- a/041-opa-integration/proposal.md
+++ b/041-opa-integration/proposal.md
@@ -29,14 +29,16 @@ Possible policies to apply could be:
 ## Support multiple policy managers
 
 Like how multiple credential managers are supported, Concourse should allow other
-policy managers than OPA.
+policy managers than OPA. Thus an interface of `PolicyCheck` should be defined, and
+OPA is one of the implementations.
 
 ## Policy check points
 
-* All API calls, for example `set-pipeline`, will also go through policy checks.
-* Because Concourse supports task file, and task file is only fetched at runtime.
-The action `set-pipeline` has no access to task files. To support policy check 
-against tasks, a policy check should be invoked before a task run starts.
+* All API calls, for example `set-pipeline`, will also go through policy checks. If
+the check doesn't pass, API should return HTTP code 403 (forbidden).
+* `UsingImage` action will be sent to OPA before Concourse launches a container in
+`check/get/put/task` steps. If the check doesn't pass, the step should return an error
+indicating policy check not pass.
 
 ## OPA configuration
 
@@ -66,14 +68,14 @@ policies to to configured in the policy engine.
 * `cluster_name`: cluster name.
 * `cluster_version`: Concourse version
 * `action`: Action name. This follows same action names used in RBAC, plus an extra
-action `RunTask`.
-* `http_method`: HTTP method of the action, for `RunTask`, this field absents.
+action `UsingImage`.
+* `http_method`: HTTP method of the action, for `UsingImage`, this field absents.
 * `user`: username who invokes the action.
 * `team`: team name.
 * `pipeline`: pipeline name. Some action is not against a pipeline, then this field 
 can be omitted.
-* `data`: For API actions, `data` should carry data from the API call; for `RunTask`
-action, data is the task configuration.
+* `data`: For API actions, `data` should carry data from the API call; for `UsingImage`
+action, data is the image configuration.
 
 For example, a policy check request against `set_pipeline` looks like:
 
@@ -106,26 +108,26 @@ to decide which actions should run policy checks, meaning that, without defining
 policy check filters, no action will run policy check.
 
 Users may not want to run policy check against all actions. For example, Concourse
-generate large amount of `ListAllPipelines`, and it makes not much sense to check
+generate a large amount of `ListAllPipelines`, and it makes not much sense to check
 it.
 
-It's assumable that users more tends to check policy against write actions than
-read-only actions. As all actions expect `RunTask` are invoked from HTTP, we 
-can provide a filter, `policy-check-filter-http-methods`, to specify HTTP 
+Users will tend to check with the policy manager for write actions rather than
+read-only actions. As all actions except `UsingImage` are invoked from HTTP, we 
+can provide a filter, `policy-check-filter-http-method`, to specify HTTP 
 methods via which actions are invoked. To skip read-only action for policy 
 check, users may set `POST,PUT,DELETE` to the filter `policy-check-filter-http-
-methods`, so that `GET` actions will not go through policy check.
+method`, so that `GET` actions will not go through policy check.
 
 User also may specifically want to or don't want to do run policy check against
 certain actions. For example, a cluster will want to check policy against only
-`SetPipeline`, or the other cluster don't want to check policy against `RunTask`,
+`SetPipeline`, or the other cluster don't want to check policy against `UsingImage`,
 for which two more filters, `policy-check-filter-action` and `policy-check-
 filter-action-skip` are supported. If an action is defined in action list, then 
 the action will always go through policy check, vice versa for action-skip list. 
 
 In summary, where are three policy check filters:
 
-* `policy-check-filter-http-methods`: defines HTTP methods of actions that
+* `policy-check-filter-http-method`: defines HTTP methods of actions that
 should go through policy checks. Default to empty list.
 * `policy-check-filter-action`: defines action names that should always go 
 through policy checks. Default to empty list.
--- a/041-opa-integration/proposal.md
+++ b/041-opa-integration/proposal.md
@@ -36,7 +36,7 @@ OPA is one of the implementations.
 
 * All API calls, for example `set-pipeline`, will also go through policy checks. If
 the check doesn't pass, API should return HTTP code 403 (forbidden).
-* `UsingImage` action will be sent to OPA before Concourse launches a container in
+* `UseImage` action will be sent to OPA before Concourse launches a container in
 `check/get/put/task` steps. If the check doesn't pass, the step should return an error
 indicating policy check not pass.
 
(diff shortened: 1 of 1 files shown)


041-opa-integration/proposal.md
```diff
@@ -0,0 +1,150 @@
...
+## Support multiple policy managers
+
+Like how multiple credential managers are supported, Concourse should allow other
+policy managers than OPA.
+
+## Policy check points
+
+* Concourse has implemented the audit function. Each action, e.g. `set-pipeline`, 
+generates an audit log. Policy checks can be done right before writing audit logs.
+So that all actions that are audited will also go through policy checks.
```
ari-becker: It seems to me that this RFC has a pre-requisite that Concourse first expand the simplistic auditing capabilities that were added in 5.2 - documenting explicitly what is audited, a JSON schema of generated audit logs, etc.

Currently, auditing is configured through boolean environment variables. Maybe this should be changed? What if a policy requires an audit which is not enabled through the environment variables - should the action automatically be denied or permitted, should it be impossible to apply the policy to begin with?

It seems to me that there should first be an RFC on maturing auditing before putting work in on policy engines.

evanchaoli: Thanks for the comment and sorry for the confusion. The policy check doesn't depend on audit at all. Let me fix the wording.

evanchaoli:
@pivotal-jamie-klassen Thanks for the suggesting regarding to naming. I have updated the doc based on your suggestion.

041-opa-integration/proposal.md
```diff
@@ -0,0 +1,147 @@
...
+When a policy manager is configured, Concourse intends to use "explicit" strategy
+to decide which actions should run policy checks, meaning that, without defining
+policy check filters, no action will run policy check.
+
+Users may not want to run policy check against all actions. For example, Concourse
+generate large amount of `ListAllPipelines`, and it makes not much sense to check
+it.
+
+It's assumable that users more tends to check policy against write actions than
+read-only actions. As all actions expect `RunTask` are invoked from HTTP, we
```
vito, line 113:
```suggestion
read-only actions. As all actions except `RunTask` are invoked from HTTP, we 
```

041-opa-integration/proposal.md
```diff
@@ -0,0 +1,147 @@
...
+User also may specifically want to or don't want to do run policy check against
+certain actions. For example, a cluster will want to check policy against only
+`SetPipeline`, or the other cluster don't want to check policy against `RunTask`,
+for which two more filters, `policy-check-filter-action` and `policy-check-
+filter-action-skip` are supported. If an action is defined in action list, then 
+the action will always go through policy check, vice versa for action-skip list. 
+
+In summary, where are three policy check filters:
+
+* `policy-check-filter-http-methods`: defines HTTP methods of actions that
```
vito, line 128: Since flags for list values are passed multiple times, one for each value, the flag name should be singular.

```suggestion
* `policy-check-filter-http-method`: defines HTTP methods of actions that
```

041-opa-integration/proposal.md
```diff
@@ -0,0 +1,147 @@
...
+
+When a policy manager is configured, Concourse intends to use "explicit" strategy
+to decide which actions should run policy checks, meaning that, without defining
+policy check filters, no action will run policy check.
+
+Users may not want to run policy check against all actions. For example, Concourse
+generate large amount of `ListAllPipelines`, and it makes not much sense to check
+it.
+
+It's assumable that users more tends to check policy against write actions than
```
vito, line 112:
```suggestion
Users will tend to check with the policy manager for write actions rather than
```

041-opa-integration/proposal.md
```diff
@@ -0,0 +1,147 @@
...
+# Proposal
+
+## Support multiple policy managers
+
+Like how multiple credential managers are supported, Concourse should allow other
+policy managers than OPA.
+
+## Policy check points
+
+* All API calls, for example `set-pipeline`, will also go through policy checks.
```
vito, line 36: What happens when an API request is rejected by the policy manager? Does the API return a certain response (status code or body)?

041-opa-integration/proposal.md
```diff
@@ -0,0 +1,147 @@
...
+* Because Concourse supports task file, and task file is only fetched at runtime.
```
vito, line 37: What happens when a task step is rejected by the policy manager? Does the step error with a message?

041-opa-integration/proposal.md
```diff
@@ -0,0 +1,147 @@
...
+* Some docker images are not not allowed.
+* and so on ...
+
+
+# Proposal
+
+## Support multiple policy managers
+
+Like how multiple credential managers are supported, Concourse should allow other
+policy managers than OPA.
```
vito, line 32: This sounds reasonable, but I think the RFC should show how this would work or leave it as an open question.

041-opa-integration/proposal.md
```diff
@@ -0,0 +1,147 @@
...
+## Policy check points
+
+* All API calls, for example `set-pipeline`, will also go through policy checks.
+* Because Concourse supports task file, and task file is only fetched at runtime.
+The action `set-pipeline` has no access to task files. To support policy check 
+against tasks, a policy check should be invoked before a task run starts.
+
+## OPA configuration
+
+* CONCOURSE_OPA_URL - URL of OPA service including path
```
vito, line 43: Do you see this as something that should only be configured cluster-wide? Or would it make sense to you to have this configured on a per-team basis?

evanchaoli: For our use cases, we don't need per-team OPA. If a team/pipeline wants to use their own OPA, an OPA resource type can be used.

041-opa-integration/proposal.md
evanchaoli: Http code 403 (forbidden) should be returned.

041-opa-integration/proposal.md
evanchaoli: Sorry for late update. I have changed `RunTask` to `UsingImage` in the implementation. Just updated the RFC. If `UsingImage` doesn't pass policy check, it should return an error `PolicyCheckNotPass` from function `Run` of step `check/get/put/task`.

041-opa-integration/proposal.md
evanchaoli: Added a description.

041-opa-integration/proposal.md
```diff
@@ -0,0 +1,149 @@
...
+
+Like how multiple credential managers are supported, Concourse should allow other
+policy managers than OPA. Thus an interface of `PolicyCheck` should be defined, and
+OPA is one of the implementations.
+
+## Policy check points
+
+* All API calls, for example `set-pipeline`, will also go through policy checks. If
+the check doesn't pass, API should return HTTP code 403 (forbidden).
+* `UsingImage` action will be sent to OPA before Concourse launches a container in
```
aoldershaw, line 39: I can also imagine wanting to prevent the use of `privileged: true` for tasks (perhaps even on a per-team basis), which also feels like it should be handled as some `PolicyCheck` (and probably OPA?).

This makes me wonder: are there any other things besides images (and `privileged`) that operators would be interested in checking against OPA? The answer may just be no, but for arguments sake, what if Concourse didn't try to be opinionated about what actions to check against, and instead supported 4 actions beyond to the API actions: `Check`, `Get`, `Put`, and `Task`. Each action's `data` was the entire step definition.

Granted, this would mean that if all you cared about was the image used, you'd need 4 identical policy check definitions - but, by it's nature, this also affords operators more flexibility in what kind of checks they perform. This may be overkill depending on the use cases in practice, though. However, it does feel more similar to the approach taken for API endpoints (being unopinionated and sending the entire body of the API call).

041-opa-integration/proposal.md
```diff
@@ -0,0 +1,149 @@
...
+* `cluster_name`: cluster name.
+* `cluster_version`: Concourse version
+* `action`: Action name. This follows same action names used in RBAC, plus an extra
+action `UsingImage`.
+* `http_method`: HTTP method of the action, for `UsingImage`, this field absents.
+* `user`: username who invokes the action.
+* `team`: team name.
+* `pipeline`: pipeline name. Some action is not against a pipeline, then this field 
+can be omitted.
+* `data`: For API actions, `data` should carry data from the API call; for `UsingImage`
```
vito, line 77: It's probably worth noting here *what* data gets sent during `UsingImage` - we've gone through a few iterations in https://github.com/concourse/concourse/pull/5034

In particular it's probably worth mentioning how credentials are handled, and the potential risk of transmitting them if a credential manager is not used or if secret redaction is not enabled.

evanchaoli: Added a note for image_source redaction.

marco-m-pix4d:
Hello @evanchaoli,
what is the status of this RFC wrt the merged code in concourse/concourse#5034?
should this RFC be merged as done or are there missing parts?

anderseknert:
Came across this by random and it looks nice! Planning to add this to the OPA ecosystem page. One thing I couldn't find on the Concourse site was docs around how to integrate with OPA. I am not too familiar with the project so might just be I'm not looking in the right places. Could anyone help me out? :)

marco-m-pix4d:
@anderseknert the only docs I am aware of are in the PR description, see https://github.com/concourse/concourse/pull/5034#issuecomment-693252774

marco-m-pix4d:
@evanchaoli ping regarding https://github.com/concourse/rfcs/pull/41#issuecomment-769054867

anderseknert:
Thanks @marco-m-pix4d 👍  Hoping that gets added to the docs then so that I can point to this as a good example of OPA in CI/CD integrations.

taylorsilva:
Merging as the basic implementation is there and is functional for users today.

PR #43, opened 2020-02-03
RFC: Tasks queue.

aledeganopix4d:
[Rendered](https://github.com/Pix4D/rfcs/blob/043-tasks_queue/043-tasks_queue/proposal.md)

# Acknowledgement
I would like to publicly thank my friend and colleague David Wagner (@wagdav) for abstracting and
summarizing the concepts exposed in this RFC.

### DIFF
--- a/043-tasks_queue/proposal.md
+++ b/043-tasks_queue/proposal.md
@@ -1,0 +1,145 @@
+# Summary
+
+We propose to enhance Concourse's task scheduling process. Instead of tasks immediately picking a compatible worker, a task is passed to a new component: the Resource Pool. 
+
+The Resource Pool decides when the execution of a task is possible (contrary to the placement strategies which decide which worker is preferred). The Resource Pool component allows the implementation of specific, gradual improvements to making Concourse's scheduling safe, fair and scalable.
+
+
+# Motivation
+
+There have been multiple issues ([#2577](https://github.com/concourse/concourse/issues/2577), [#2928](https://github.com/concourse/concourse/issues/2928) etc) opened regarding Concourse's build/container scheduling and the container placement strategies currently present (`random`, `volume-locality`, `fewest-build-containers` and `limit-active-task`).
+
+As Concourse operators, today we see that Concourse's scheduling process is:
+
+1. not safe: workers are not protected from overload. When the system is overloaded it has no chance to recover.
+2. not fair: tasks are racing for being executed without any specific order
+3. not scalable: it is not possible to measure if the number of workers is appropriate to the current user demand.
+
+To highlight the problem in more detail here's an overview of Concourse's task scheduling process.
+
+1. the module scheduler selects the tasks that can be run (all their inputs are available, their job/pipeline is not paused)
+2. the task submits itself to the WorkerPool
+3. the WorkerPool selects a compatible worker according to the configured PlacementStrategy
+4. the task runs
+
+During this process no component asks whether the system, in particular the workers, are capable of accepting more tasks. The different placement strategy implementations only express preferences.
+
+The scheduling process requires a new component which keeps track of the available resources as tasks start, execute and stop on the available workers.
+
+
+# Proposal
+
+Let's take a look at the RunTaskStep function from (`atc/worker/client.go`)
+
+```go
+func (client *client) RunTaskStep(...) TaskResult {
+    //...
+    chosenWorker, err := client.chooseTaskWorker(...)
+    //...
+}
+```
+This function is called from the `Run()` method of TaskStep, which is effectively responsible for starting the task's container.
+
+The task chooses a worker given the placement strategy and the worker starts executing the corresponding container. This happens without taking into account the available resources on the worker. If users submit more jobs than the workers can instantaneously handle, the system becomes overloaded. Additionally, in case of overload no corrective action is possible since the tasks cannot re-schedule themselves.
+
+Instead of directly allocating a worker from a task, we propose the following, extended worker allocation process:
+
+![resource pool schema](https://ci-pix4d-internal-static.s3-eu-west-1.amazonaws.com/concourse-scheduling-rfc.png)
+
+A new component, called Resource Pool is placed between the worker client (the ATC) and worker pool. The client would interact with the worker pool as follows:
+
+1. The client requests a worker from Resource Pool.  This call blocks until a worker becomes free.
+2. The resource pool returns the worker to the client.  The worker has been chosen from the available workers using the container placement strategy.
+3. The client starts the task container on the allocated worker
+4. The task container stops
+5. The worker client releases the worker's resources
+
+The key idea is the resource pool throttles the starting of containers and protects the workers from overload.
+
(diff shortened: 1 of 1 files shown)


ari-becker:
Just to clarify - you see this RFC as being relevant when Kubernetes is not being used as a runtime? Or alongside using Kubernetes as a runtime for e.g. OS X tasks (as you referenced in your [comment on the Kubernetes RFC](https://github.com/concourse/rfcs/pull/22#issuecomment-511309425))?

aledeganopix4d:
I think this is mostly relevant when K8s is not (yet) used as scheduler. Especially given that the timeline of this RFC should be shorter than the k8s integration (maybe?).

wagdav:
Hello, would somebody please look at this RFC and comment what you think about it? Thanks!

vito:
@wagdav Sorry this hasn't gotten any attention from us, v6.0 is dragging on so I'm trying to focus on getting that out the door. Should be some time this week.

vito:
I haven't forgotten! 😅 I'm working on an RFC round-up blog post now and I'm going to include this one in it so it can get more eyes on it. Thanks for your patience!

043-tasks_queue/proposal.md
```diff
@@ -0,0 +1,145 @@
...
+
+```go
+func (client *client) RunTaskStep(...) TaskResult {
+    //...
+
+    requiredResources := make(map[string]int)
+    requiredResources["active-task"] = 1
+ 
+    chosenWorker := resourcePool.RequestWorker(requiredResoures, placementStrategy, taskParameters...)
+    defer resourcePool.ReleaseWorker(requiredREsources, chosenWorker)
```
vito, line 90: Although this is captured as a `defer`, it's probably worth reasoning about the doomsday scenario of an ATC getting `kill -9`ed. Would these requests leak? Do we need some sort of GC semantics for them? 🤔

aledeganopix4d: IIRC when the ATC starts again (or a different ATC for that matter), it will go through this code again, then if the task is still being executed it will wait for it, otherwise will exit immediately, calling the defer.

This is already the case for the placement strategy `limit-active-tasks` where we indeed tested that when the ATC recovers all the "flying" tasks also releases the resources no longer in use.

manuel-pix4d:
Hello @vito, it seems that this RFC turn in sleeping mode (was there in the first RFC round-up but disappear in the other RFCs). Is it possible to get a feedback around this RFC ?

vito:
@manuel-pix4d I'll include it in the next one!

043-tasks_queue/proposal.md
```diff
@@ -0,0 +1,145 @@
+# Summary
+
+We propose to enhance Concourse's task scheduling process. Instead of tasks immediately picking a compatible worker, a task is passed to a new component: the Resource Pool.
```
jchesterpivotal, line 3: I feel like "Resource Pool" is slightly misleading. It seems like what is happening is that ready-to-run tasks are being buffered until resources are available. "Resource Pool" to me sounds more worker-centric than task-centric.

043-tasks_queue/proposal.md
```diff
@@ -0,0 +1,145 @@
...
+
+![resource pool schema](https://ci-pix4d-internal-static.s3-eu-west-1.amazonaws.com/concourse-scheduling-rfc.png)
+
+A new component, called Resource Pool is placed between the worker client (the ATC) and worker pool. The client would interact with the worker pool as follows:
+
+1. The client requests a worker from Resource Pool.  This call blocks until a worker becomes free.
+2. The resource pool returns the worker to the client.  The worker has been chosen from the available workers using the container placement strategy.
+3. The client starts the task container on the allocated worker
+4. The task container stops
+5. The worker client releases the worker's resources
```
jchesterpivotal, line 55: I would rather that Workers "pull" from tasks that are ready to run, rather than the task itself blocking on finding a worker.

I guess this is why the name is "Resource Pool", not "Tasks Pool".

aledeganopix4d: This would be a lot more complex than it sound: currently the workers (in my understanding at least), are quite blind about the tasks and resources that they are required to process.
So first of all we would need to make the workers aware of the workload they have to process and then let them manage it.
That solution would also make it a lot harder to have meaningful metrics about the tasks (those waiting and those running).
In the end I don't think there is really nothing to gain from the "pull" architecture as the ATC will always be there as the central dispatch and point of contact of the DB where, in any case, the state of all tasks must be recorded.

043-tasks_queue/proposal.md
aledeganopix4d: Yes, in this instance the "resource" is the worker (alas "workers pool" is already in use in the code base).

kcmannem:
```
WorkerAccess.Request()
WorkerAccess.Release()
```
🤷‍♂️ 

Do agree that resource is already overloaded

043-tasks_queue/proposal.md
```diff
@@ -0,0 +1,145 @@
...
+* As resources become available they are offered to the first task in the queue.
+
+This resource pool implementation would be independent of the chosen worker placement strategy. Thus all worker placement strategies benefit from a safe and fair task distribution process.
+
+# Open Questions
+
+* Is "Resource Pool" a good name? (resource means something different in the context of pipeline jobs and tasks)
+* Could the functionality of the "Resource Pool"' be fit some already existing component?
+* Should this be tasks scheduling or more generic step scheduling? That is to include check containers (resource-type checks etc.)
+* What is the relationship to the new LIDAR component?
```
clarafu, line 114: There shouldn't be any more complexity with the new lidar component, it is just a different way of scheduling checks as the checks are queued and run as a step https://github.com/concourse/concourse/blob/master/atc/exec/check_step.go. If we wanted to later limit the number of checks run on a worker (not sure if we would ever want that though because checks should be incredibly quick to run), we could add the limiting functionality to the resource pool when we fetch the worker to run the check step on.

aledeganopix4d:
Hello, since https://github.com/concourse/concourse/pull/5981 has been merged, my team and I have implemented a different strategy to ensure tasks never wait too long in the queue: we are now able to autoscale our workers quite reactively to meet the demand and thus the fact that the tasks waiting to be executed are no longer in a queue is not a big problem for us anymore.
Indeed we observed that no task waits longer than ~15m (which correlates with the startup time of our workers), regardless of when it gets executed with respect to the other tasks waiting.

For this reason and for full transparency on our side I'd like to notify that even if this RFC will ever be approved we would not be able to work on it as it wouldn't bring value to our team anymore.

Should we close it?

vito:
@aledeganopix4d Hmm, yeah maybe we should just close it. Now that we're actually starting on the K8s runtime work there's not much appetite for spending mental cycles on this in parallel.

I do think it's important for the non-K8s runtime experience to be better, but maybe someone can revive this sometime in the future. Who knows, maybe we'll have completely different ideas, or a better architecture to build upon.

Thanks again for putting this together, and sorry that it kind of fell through the cracks and we weren't able to engage on it as much as I'd like to have been able to.

PR #44, opened 2020-02-27
RFC: Pipeline Kubernetes CRD

cirocosta:
[Rendered](https://github.com/concourse/rfcs/blob/044-api-kubernetes-crs/044-api-kubernetes-crs/proposal.md)


muntac:
This looks cool! I have two comments. 

1. This might be more relevant for a V2 of this controller. Do you think it would be useful to have the ability to parameterize the pipeline resource as well so we can have different variations of the same pipeline come up?

I don't know how many users run into this issue but on a past team we had a several variations of the same pipeline with slightly different resource parameters. Whenever we deployed a fresh Concourse we found ourselves desiring a way to bring up all those pipelines automatically.

2. Not sure if this was already implied but it would be nice to be able to use the controller to also "export" all the current pipeline definitions. These could then be imported by the controller into a new deployment. This would also for the time being serve as a partial workaround for the parameterization case.

jchesterpivotal:
As it so happens, @topherbullock and I built [a quick spike](https://github.com/jchesterpivotal/knative-build-pipeline-poc) of this concept back in 2018.

The simplest hack was to directly pass through the body of the pipeline as a `runtime.RawExtension`, which you can see in [`pipeline_types.go`](https://github.com/jchesterpivotal/knative-build-pipeline-poc/blob/master/pkg/apis/concourse/v5alpha1/pipeline_types.go).

andy-paine:
This might be of interest at least with respect to user experience https://github.com/alphagov/gsp/tree/master/components/concourse-operator

taylorsilva:
No plans to do this now.

PR #45, opened 2020-03-04
RFC: Concourse Kubernetes Operator

cirocosta:
[Rendered](https://github.com/concourse/rfcs/blob/045-kubernetes-operator/045-kubernetes-operator/proposal.md)


045-kubernetes-operator/proposal.md
```diff
@@ -0,0 +1,315 @@
...
+object that represents a Concourse installation:
+
+
+```yaml
+apiVersion: concourse.concourse-ci.org/v1
+kind: Concourse
+metadata:
+  name: hush-house
+spec:
+  replicas: 2
```
aoldershaw: Am I correct to infer that this CRD only represents ATC's, and isn't responsible for creating workers? Or is this looking to a point beyond having worker pods, where the cluster itself is a "worker"?

I'm not too familiar with how the [k8s worker experiment](https://github.com/concourse/concourse/pull/5223) you created works, but it may be the case that with "k8s workers", the ATC creates pods for steps/resource checks (and there's no dedicated "worker" pods)?

taylorsilva:
One blocker for this is that the core team has no desire to build or maintain an operator right now. I don't see the team trying to support anything k8s-related besides the chart right now.

This talk also got me wondering if writing an operator is even a good idea for Concourse: https://www.youtube.com/watch?v=CwVCfl4qpdg

- Secret generation is really rough on k8s right now. I think this is a problem the platform needs to solve. The operator solution here feels like a stopgap until the platform catches up
- The motivation for k8s Operators seems to be for stateful applications. Concourse itself is a pretty good example of a classic [12-factor app](https://12factor.net/). All state is in the db which users are probably not using the postgres chart for.

Postponing for now, we'll see how the k8s landscape changes in the future.

rhoughton-pivot:
The k8s runtime for Concourse feels like a natural fit. If a team correctly sizes (resource.limits) a given `step` or `task` then k8s does scheduling and scaling. 

Sad to see this delayed.

pratikbin:
> Postponing for now, we'll see how the k8s landscape changes in the future.

Agree. I don't see a need for concourse operator right now, There are other proposals that should be prioritized.

rkoster:
One other way to go about this issue is to use Carvel.dev and its [secret-gen-controller](https://github.com/vmware-tanzu/carvel-secretgen-controller). This is the approach that has been taken by the [reusable concourse templates used by the bosh community][https://github.com/cloudfoundry/bosh-community-stemcell-ci-infra/].

DrummyFloyd:
any news ? 
because natively , i would think taht concourse goes weel with concourse.
it has the quite same philosophy.

would be cool to have this features.

like a plugins features.>
like this the maintener of the RFC can maintains this and be help by the community.
plus Cconourse team did not need to mains something like this.

if you don't want such a thing. even if i love the idea of an operator.

PR #47, opened 2020-03-23
RFC: allow running web backend and frontend in separate processes

jamieklassen:
[Rendered](https://github.com/pivotal-jamie-klassen/rfcs/blob/microservices/47-backend-frontend/proposal.md)

### DIFF
--- a/47-backend-frontend/proposal.md
+++ b/47-backend-frontend/proposal.md
@@ -43,30 +43,23 @@ I think there's a lot of value in placing control over these things in the hands
 
 # Proposal
 
-Broadly, we need a solution that lets operators choose whether they want to run the various Concourse computations in `ifrit/grouper.Members` or actual separate unix processes -- these things are supposed to be nearly-interchangeable concepts anyway.
+Broadly, we need a solution that lets operators choose whether they want to run the various Concourse computations in `ifrit/grouper.Members` or actual separate unix processes -- these things are supposed to be nearly-interchangeable concepts anyway. We could provide this configurability at many levels of granularity, so let's focus this RFC on a simple case.
 
-## More subcommands
+## Frontend and Backend
 
-Concretely, I want finer-grained `concourse` subcommands besides `quickstart`, `web` and `worker`. I want a bit more - if we look through the past at events where a new "member" was added, I want to know that there will always be a sub-command for it.
+Concretely, I want finer-grained `concourse` subcommands besides `quickstart`, `web` and `worker`. Let's add two more:
 
-As an initial iteration, why not the following:
-* `concourse ui` for the frontend, web handlers
-* `concourse api` for the http and https servers
+* `concourse frontend` for the http and https servers serving the API and web UI
 * `concourse backend` for the scheduler/gc/logcollector/syslogdrainer/etc
-* `concourse gateway` for the TSA
 
-## More flags
+Ideally each of these subcommands should really only be concerned with its own configuration - i.e. there's no need to configure a gc interval on the frontend. However, they will have some overlapping configuration (like the database) that will need to be duplicated.
 
-Another option, rather than aggressively creating these subcommands , we could add flags to enable/disable each one. Then operators could add a `DISABLE_*` env var when they suspected a component might be misbehaving, or they could set up deployment manifests where each member was running in a separate process/container by carefully toggling all the members but one at a time.
+Furthermore, `concourse web` should continue to behave the same as it does now. 
+
+This would coarsely solve the problem posed in the above narrative - we could have quickly diagnosed that the memory consumption was coming from the `frontend` process.
 
 # Open Questions
-
-I know, it's basically "microservices with a shared database", but that's pretty much what our architecture has already been anyway.
-
-Maybe this approach could lead to each member maintaining its own persistence? Moving towards a (hopefully more effective) distributed persistence solution? If we knew that there was a strongish decoupling between the build event server and the rest of the system (like it was the only one that did reads on the `build_events` table), maybe this would open the possibility of configuring a different storage backend (like a blobstore or a [second postgres](https://github.com/concourse/concourse/issues/5306) or what have you)
 
 # Answered Questions
 
 # New Implications
-
-Do we go all-in on microservices accoutrements? Service discovery? A message bus? the 'components' table is pretty much a service registry, as far as I understand the concept. the TSA is a decent choice for a service registry API server, since it's already expected that it will be publicly reachable for worker registration.
--- a/47-microservices/proposal.md
+++ b/47-microservices/proposal.md
@@ -1,0 +1,72 @@
+# Summary
+
+Let's create a way to distribute the various procedures that the web node runs across multiple unix processes.
+
+# Motivation
+
+## A story of observability pain
+
+Recently I was required to debug a Concourse cluster, in production, that was having essentially a total outage: The web nodes were no longer serving API requests. They were completely maxing out their memory consumption to the point that they failed their health checks and got restarted by monit (this was a bosh deployment).
+
+I was on a long phone call with some stressed-out operators and everybody was somewhat-aimlessly brainstorming. At one point, somebody said
+
+> it looks like there's a regular pattern of steps - the memory shoots up, stays at max for a minute or so, falls down to zero and then shoots up within about another minute. This doesn't look like a leak, but more like a resource-consumption bug in some kind of periodic asynchronous procedure.
+
+She could definitely be right. In my heightened state (I was trying to fix a production issue in real-time over a Zoom call) I started rapidly reading through the codebase - what are all the periodic asynchronous procedures that the web nodes run?
+
+I was looking through `atc/atccmd/command.go` and realized that the things I was looking for ("periodic asynchronous procedures") all fit more precisely into the category of `github.com/tedsuo/ifrit/grouper.Member` values.
+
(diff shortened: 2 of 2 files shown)


pivotal-jwinters:
I love the new subcommand proposal. I've been wanting to separate these for a while, thanks for formalizing these ideas.

One of the drivers for https://github.com/concourse/concourse/pull/5305 was making gc easier to configure so that its easier to separate out.

I'm not sure about the `ENABLE_*` flags. This was supposed to have been addressed with the `components` table, where each component can be paused without having to restart the web node. Is there something that the components table doesn't address?

47-microservices/proposal.md
```diff
@@ -0,0 +1,72 @@
...
+
+## More subcommands
+
+Concretely, I want finer-grained `concourse` subcommands besides `quickstart`, `web` and `worker`. I want a bit more - if we look through the past at events where a new "member" was added, I want to know that there will always be a sub-command for it.
+
+As an initial iteration, why not the following:
+* `concourse ui` for the frontend, web handlers
+* `concourse api` for the http and https servers
+* `concourse backend` for the scheduler/gc/logcollector/syslogdrainer/etc
+* `concourse gateway` for the TSA
```
pivotal-jwinters: It would also be nice to see a separate `concourse auth` component.

pivotal-jwinters:
Also I've actually attempted a spike to separate the `ui` from the `api` more than once. There are a couple problems:

1) We would either need to support CORS or use a web proxy so that web requests appear to originate from the same origin.

2) So many ports. Currently operators only need to expose a single port for all of concourse. In the microservice world they either need to expose many ports, or more proxies.

pivotal-jamie-klassen:
I'm starting to have serious reservations about the use of the term 'microservices' to describe this change. The key idea here is to improve observability of the web node. The proposed solution happens to leverage existing mechanisms from Linux by allowing operators to configure the web node to spread its work across multiple co-located processes. Microservices, by definition, are 'independently deployable services modeled around a business domain'. As is perhaps typical for discussions involving microservices, that is a _way_ bigger change than I'm trying to suggest - I'm not trying to make any recommendations about deployment cycles or business domains or separating the persistence layer or distributed transactions etc etc etc. This is not really an architectural suggestion at all.

pivotal-jwinters:
Given the difficulties of splitting ui/api, I think we can get some big wins, just by splitting up the atc into frontend (api, web, auth) and backend (gc, scheduler, lidar, etc.). Basically the current `atc run` command becomes `frontend` and `backend`.

                                    +------------- quickstart -------------+
                                    |                                                                    |
           +----------  web ----------+                                      worker
            |                                               |
          tsa                       +---------- atc ----+---------------+
                                       |                                  |                            |
                                 frontend                   backend               migrate

Thoughts?

pivotal-jamie-klassen:
The key acceptance criteria is this - I want to know which API endpoint is eating all my memory without pulling a heap dump. Thinking lean here, suppose we just had a way to run the `ListAllJobs` function in its own process, with the guarantee that this process will handle all `/api/v1/jobs` requests. What consequences does this entail?

unix processes each have to have their own file handles - the web node has three kinds of file handles of interest that I can think of: stdout/stderr for logs, the ones that point to config files, which I believe should be opened, read and then closed as part of the startup process; and TCP sockets. What kind of sockets does it have?

* listeners:
    * HTTP (for web & api)
    * HTTPS (ditto)
    * SSH (TSA)
    * pprof debug
    * forwarded worker connections
* database connections

The key consequence that stands out to me is what @pivotal-jwinters has pointed out: Which process actually serves the traffic, i.e. owns the HTTP listener? While we're at it, we probably want a consistent notion of a logical database connection pool - who will coordinate and make sure that the set of connections is equitably distributed between these processes?

When we ask the question this way, the solution that suggests itself is to rely primarily on proxies: 

* a reverse proxy receiving all the web traffic and somehow 'routing' the 'work' of serving those requests to the correct process
* a db proxy process on the other end that handles the logical pooling.

We can also turn this question on its head. Let's assume that the web node continues its usual existence and manages all the interesting file handles - then how do we define the inter-process communication between the web process and the all-jobs-lister process? The first design that comes to mind when I ask the question this way would be to use the most-minimal unix IPC mechanisms - the web process is a parent of the all-jobs-lister, which it spins up on-demand, passing any information in the request via command-line parameters or environment variables, and when the all-jobs-lister terminates, the web receives the information it needs from an exit code or perhaps by reading stdout.

cirocosta:
Hey @pivotal-jamie-klassen, thanks for bringing this up!

I wanted to challange a point with regards to the key problem (as you
mentioned, "improving **obersability** of the web node"):

> I want to know which API endpoint is eating all my memory without pulling a
> heap dump. 
>
> (from https://github.com/concourse/rfcs/pull/47#issuecomment-607928996)

Could you expand on why relying on heap dumps is undesirable?

47-backend-frontend/proposal.md
```diff
@@ -0,0 +1,65 @@
...
+https://github.com/concourse/concourse/blob/release/5.5.x/atc/atccmd/command.go#L11
+
+and then we discovered that the `ListAllJobs` endpoint was ultimately mostly responsible for the memory consumption, and we made a PR adding a config flag that to disable specifically that badly-behaved endpoint: https://github.com/concourse/concourse/pull/5340
+
+## A request for non-invasive operator-configurable observability
+
+How can we improve situations like this in the future? I would like the following acceptance criteria:
+
+A way to report compute-resource consumption of Concourse by "member" (or smaller), such that
+* data can be harvested in real-time
```
cirocosta, line 34: Could you expand on what did you mean by this? Do you mean that that should be continuously collected throughout the execution of the member, or that at the moment I need, I should be able to retrieve it (i.e., take a snapshot)?

pivotal-jamie-klassen:
@cirocosta two benefits:
* operators don't need to learn how to use pprof to reason about their systems
* process-level metrics can be measured in real-time (like watching `top`, etc)

my hope is that by having the option to have the measurements in real time, overall anxiety is reduced - a big part of my story is frantically searching through code and that wasn't very fun. I'm not sure I can draw a strong link between "real-time" and "reduced anxiety" beyond the fact that "big enterprises like graphs".

47-backend-frontend/proposal.md
```diff
@@ -0,0 +1,65 @@
...
+
+I can imagine a world where a common way to introduce Concourse features is by adding or removing "members" of this kind - frankly, `lidar`, the big feature in v5.6.0, was basically a drop-in replacement for the existing `radar` that performed resource checks.
+
+I think there's a lot of value in placing control over these things in the hands of the operator, especially when operators may have strict change management controls that forbid them from modifying the actual software in production, but may be more lenient about making configuration changes.
+
+# Proposal
+
+Broadly, we need a solution that lets operators choose whether they want to run the various Concourse computations in `ifrit/grouper.Members` or actual separate unix processes -- these things are supposed to be nearly-interchangeable concepts anyway. We could provide this configurability at many levels of granularity, so let's focus this RFC on a simple case.
```
cirocosta, line 46: I'm struggling a bit with understanding if I'm looking at the RFC with the right angle :grimacing: 

e.g., if I take this exactly sentence, and append to it a "so it solves what I'm trying to solve", is the following correct?

> [...] we need a solution that lets operators choose whether they want to run the various Concourse computations in `ifrit/grouper.Members` or actual separate unix processes, **so that observability of the web node is improved, being even able to know which API endpoint is eating all my memory without pulling a heap dump** 

(just trying to make sure I'm with the right mindset)

pivotal-jamie-klassen: I've definitely made a mess here - as though talking about microservices didn't give it away. I should maybe restate my goal as "figure out WTF is eating my memory without a heap dump". In this case the RFC just proposes a way to figure out whether the frontend or backend is to blame.

pivotal-jamie-klassen: The key thing that is also missing is the ~3h offline conversation between @pivotal-jwinters and I where these goals were heavily revised...

ringods:
I perceive this RFC as a mixture of a problem description and solution to narrow but not solve the problem.

The problem in the narrative is one of observability as the title mentions: you want operational insight in what is behaving wrong. 

While not wanting to promote this tool (not affiliated BTW), may I point everybody to the [Honeycomb blog](https://www.honeycomb.io/tracing-blog/)? A number of their tracing articles describe the need for such insights at run time and define the problem as follow: find out about your unknown unknowns.

Reading further in the RFC, the proposal goes immediately into chopping up the process in smaller pieces. While this narrows the area of the code you have to look into, you still have the same problem albeit on a smaller scale.

I wanted to make a counterproposal to integrate [OpenTelemetry]() in the code base, but I was surprised to see that it is already in the Go dependencies:

https://github.com/concourse/concourse/blob/5d47620249f15478765e23075fc3576cd102812c/go.mod#L103-L105

I'm interested to hear if the OpenTelemetry API usage is also fully integrated in the code base. If not, I would first make work of that as it will bring you the needed data to guide you in deciding what should best be split of as separate processes.

Stop logging, start tracing! 😉

cirocosta:
Reading https://github.com/concourse/concourse/issues/5360#issuecomment-608997452

> The web node CPU utilization looks like it's taken load off one instance, but not the other. Not sure how to interpret this. 

reminded me of this RFC - having `web` split into discrete pieces could indeed help (is most of the work coming from scheduling? is it from serving api requests? scanning for checks?), even though ultimately root-causing would only be truly possible via profiling

PR #48, opened 2020-04-14
RFC: Standardize API error messages

taylorsilva:
[Rendered](https://github.com/taylorsilva/rfcs/blob/standard-api-error-messages/048-standard-error-messages/proposal.md)

Co-authored-by: Taylor Silva <tsilva@pivotal.io>

### DIFF
--- a/048-standard-error-messages/proposal.md
+++ b/048-standard-error-messages/proposal.md
@@ -37,7 +37,7 @@ to the JSON object.
 
 # Open Questions
 
-_Couldn't think of any_
+* What format should we use? e.g. There's the [jsonAPI spec](https://jsonapi.org/format/#errors)
 
 # Answered Questions
 
--- a/048-standard-error-messages/proposal.md
+++ b/048-standard-error-messages/proposal.md
@@ -1,0 +1,49 @@
+# Summary
+
+This RFC proposes that the Concourse API standardize on returning human and
+machine readable error messages.
+
+Authors: @taylorsilva @aoldershaw
+
+# Motivation
+
+As a consumer of the Concourse API (e.g. `fly`) I want to return a helpful error
+messages to my end-user. This proves to be difficult to do because all I can do
+is imprecisely reason about why an error occured based on the HTTP status code
+returned by the API.
+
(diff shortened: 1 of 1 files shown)


048-standard-error-messages/proposal.md
```diff
@@ -0,0 +1,49 @@
...
+}
+```
+
+If further arbitrary details are required any number of extra keys can be added
+to the JSON object.
+
+
+# Open Questions
+
+_Couldn't think of any_
```
cirocosta, line 40: Perhaps we could include a question about which format to use? e.g., there's https://jsonapi.org/format/#errors, but I guess there are other formats too that are probably more widespread? 

I remember @pivotal-jwinters  was looking at various things related to a "concourse API V2", and even though I don't remember a discussion specifically about this, he might've stumbled upon formats widely used out there?

taylorsilva: We were looking for something exactly like that! Our searches failed us 😞 

All the examples we found were similar to that jsonapi spec, but every example we saw was pretty much doing its own thing.

048-standard-error-messages/proposal.md
```diff
@@ -0,0 +1,49 @@
...
+
+Example: The API endpoint for getting a pipeline config may return HTTP 404 in
+two situations: 1) the config is not found and 2) the pipeline is archived. It's
+currently not possible without further introspecition to distinguish between
+these two cases at the client level.
+
+
+# Proposal
+
+When the Concourse API returns an error (4xx and 5xx status codes) it should
```
chenbh, line 23: Should we differentiate between different levels of verbosity for 4xx vs 5xx codes? 4xx codes usually indicate user error and misuse of the API, in which case a helpful error message would be appreciated. 5xx codes indicate something has gone wrong on the server side and there's nothing the user can really do with the error message.

chenbh (commented):
I'm not sure this qualifies as a RFC under then new guidelines, it feels like this falls under the `Small features which only impact a narrow use case and affect users in an obvious way.` category. wdyt?

PR #53, opened 2020-05-10
RFC: Configurable Build Event Stores

aoldershaw:
[Rendered](https://github.com/aoldershaw/rfcs/blob/config-event-store/053-configurable-event-store/proposal.md)

Experimental PR: https://github.com/concourse/concourse/pull/5651


053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,450 @@
...
+
+# Answered Questions
+
+
+# New Implications
+
+* By default, there will be no change - Concourse will default to using a
+  Postgres `EventStore` using the same main Postgres database.
+* Operators will be able to configure a different store
+* If they configure a new store, they won't be able to access any existing
```
evanchaoli, line 448: I think this feature has to provide a migration solution, otherwise production clusters would be hard to move on external storage.

evanchaoli:
Awesome!

053-configurable-event-store/proposal.md
aoldershaw: Yeah I agree. I'd love to hear any thoughts you have on the best way to approach this. I can think of a couple options:

1. Introduce a `concourse migrate-build-events` command that migrates everything over to the configured build event store. It could possibly default to migrating from the Postgres database to whatever event store is configured, but it could also take a separate `from` configuration in case you want to migrate from a non-Postgres backend to a new backend.

2. Migrate out of Postgres "on the fly". i.e. we could have a component like the build reaper that migrates old builds to the new store, a few at a time. In this case, we'd probably still want some way to view build logs from Postgres in case they haven't yet been migrated, for which we could construct a `FallbackEventStore` that does something like this:

```go
type FallbackEventStore struct {
    EventStore
    Fallback EventStore
}

func (f *FallbackEventStore) Get(build db.Build, requested int, cursor *Key) ([]event.Envelope, error) {
    // first, try the main event store:
    events, err := f.EventStore.Get(build, requested, cursor)
    if err == nil && len(events) > 0 {
        return events, nil
    }
    // the build wasn't found (or at least has no events in the main store) - so try the Fallback (Postgres)
    // if this also errors, we'll probably want to return the err from the primary EventStore, but you get the idea
    return f.Fallback.Get(build, requested, cursor)
}
```

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+      }
+  }
+  ```
+
+  However, when possible, consumers of these build events should prefer to use
+  `db.Build.Events()`, which queries the `EventStore` and subscribes to a
+  notification bus to allow tracking in-progress builds (more on this in
+  [Notifications](#notifications)).
+
+  `Get` also takes in `requested` number of events to fetch. This is a soft
```
cirocosta, line 143: pls bear with me, but I got a bit confused with the wording here :sweat_smile:  so, is `requested` meant to be the maximum number of events that would come back, regardless of how many events are there?

e.g., a [`read(2)`](http://man7.org/linux/man-pages/man2/read.2.html) with a `count = 1024` means that it'll never try to fill my buffer with more than `1024` bytes - is this what you're trying to achieve?

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+* `DeletePipeline(pipeline db.Pipeline)` deletes all of the builds for a given
+  pipeline. While this could be done by doing something like:
+
+  ```go
+  builds, _ := pipeline.Builds()
+  eventStore.Delete(builds)
+  ```
+
+  ...this doesn't work so well for the Postgres implementation. Currently, when
+  we delete a Pipeline from the database, `TRIGGER`s will `DROP` the
```
cirocosta, line 166: we could still `drop` even though not a from a trigger, right? I mean, from a regular `Exec`.

(you probably noticed, but, just for transparency: these are the triggers we currently have:

```
 table_schema | table_name | trigger_schema |             trigger_name             | event  | activation | condition |              definition
--------------+------------+----------------+--------------------------------------+--------+------------+-----------+---------------------------------------
 public       | pipelines  | public         | pipeline_build_events_delete_trigger | DELETE | AFTER      |           | EXECUTE FUNCTION on_pipeline_delete()
 public       | pipelines  | public         | pipeline_build_events_insert_trigger | INSERT | AFTER      |           | EXECUTE FUNCTION on_pipeline_insert()
 public       | teams      | public         | team_build_events_delete_trigger     | DELETE | AFTER      |           | EXECUTE FUNCTION on_team_delete()
 public       | teams      | public         | team_build_events_insert_trigger     | INSERT | AFTER      |           | EXECUTE FUNCTION on_team_insert()
```
)

where

```
on_team_delete
    EXECUTE format('DROP TABLE IF EXISTS team_build_events_%s', OLD.id);RETURN NULL;


on_team_insert
  EXECUTE format('CREATE TABLE IF NOT EXISTS team_build_events_%s () INHERITS (build_events)', NEW.id);RETURN NULL;


on_pipeline_delete
  EXECUTE format('DROP TABLE IF EXISTS pipeline_build_events_%s', OLD.id);RETURN NULL;


on_pipeline_insert
  EXECUTE format('CREATE TABLE IF NOT EXISTS pipeline_build_events_%s () 
        INHERITS (build_events)', NEW.id);
  
  EXECUTE format(
        'CREATE INDEX IF NOT EXISTS pipeline_build_events_%s_build_id 
            ON pipeline_build_events_%s (build_id)', NEW.id, NEW.id);

  EXECUTE format(
         'CREATE UNIQUE INDEX IF NOT EXISTS pipeline_build_events_%s_build_id_event_id 
           ON pipeline_build_events_%s (build_id, event_id)', NEW.id, NEW.id);

   RETURN NULL;
```

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+
+To support having custom event stores, introduce an `EventStore` interface:
+
+```go
+type Key interface{}
+
+type EventStore interface {
+    Setup() error
+
+    Initialize(build db.Build) error
```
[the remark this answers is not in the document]
aoldershaw: Yep should definitely be doing that 🤦

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+    Finalize(build db.Build) error
+
+    Put(build db.Build, events []atc.Event) error
+    Get(build db.Build, requested int, cursor *Key) ([]event.Envelope, error)
+
+    Delete(builds []db.Build) error
+    DeletePipeline(pipeline db.Pipeline) error
```
cirocosta, line 74: it got me wondering if we should really have these :thinking: from what I understand, the distinction between `Delete` and `Delete(Pipeline|Team)` is simply that the the one implementing the interface could do so in a more performant manner. Buut .. the Pipeline and Team concepts seem to be *too* Concourse specific, kinda "leaking" into the interface here.

tbh, I don't have a suggestion :sweat_smile: just seemed like overloading the interface a bit

053-configurable-event-store/proposal.md
aoldershaw: Yeah, I was kinda struggling with how to word this.

The intention was actually the opposite of `read(2)`. The use case I was trying to account for that sometimes it's difficult to fetch exactly `requested` events from the store. Suppose  we have a Cold Storage implementation of `EventStore` (all events for a build are stored in a single file). If `requested = 1000` and there are 1500 events for the build, it doesn't really make sense to make two `Gets` and have to download the file from Cold Storage twice. Even though the client only `requested` 1000 events, since we've already fetched all 1500, we might as well give them everything.

Granted, another way to approach this would be for the Cold Storage implementation to only return 1000 events, but keep track of the remaining 500 events for the next time `Get` is called so we don't have to download the file again. Not sure what's approach is better.

I agree that with `read(2)`, knowing that you'll never try to fill a buffer with more than `count` bytes is a good guarantee, but given that the client isn't providing a buffer to fill, my thinking was exceeding `requested` isn't dangerous (let me know if you disagree!)

053-configurable-event-store/proposal.md
aoldershaw: Yep! Again, didn't really explain it so well, but that's the intention of having the `DeletePipeline` and `DeleteTeam` methods (rather than just having a general `Delete` method that acts on builds)

So, `DeletePipeline` would call `DROP pipeline_build_events_{id}`, and `DeleteTeam` would call `DROP team_build_events_{id}` and `DROP pipeline_build_events{id}` for each pipeline in the team

053-configurable-event-store/proposal.md
aoldershaw: Yeah I agree, but couldn't really think of a better way to do it without losing the ability to `DROP` tables...

We *could* try to be smart with the Postgres implementation of the `EventStore`, where if we call `Delete(builds)` where `builds` exactly matches the list of pipeline builds, drop the table instead - though I feel like this could be finicky.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+
+func (s *SyslogForwardingEventStore) Put(build db.Build, events []atc.Event) error {
+    if err := s.syslogForwardEvents(build, events); err != nil {
+        return err
+    }
+    return s.EventStore.Put(build, event)
+}
+```
+
+How about migrating build events to cold-storage after they're completed:
```
cirocosta, line 234: that's a pretty useful use case, specially for those business that require keeping logs for .. years.

assuming this would be implemented by, say, aws' s3, one can even place lifecycle policies to their bucket (https://docs.aws.amazon.com/AmazonS3/latest/dev/lifecycle-transition-general-considerations.html) and then get some sweet price reduction automatically for those super old build events

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+
+# Proposal
+
+To support having custom event stores, introduce an `EventStore` interface:
+
+```go
+type Key interface{}
+
+type EventStore interface {
+    Setup() error
```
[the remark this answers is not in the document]
aoldershaw: Yup good call!

053-configurable-event-store/proposal.md
aoldershaw: So, to summarize, `requested` isn't an upper limit on the number of events to return.

If there are at least `requested` events in the event store for a particular build, then the implementation can decide whether it makes sense to return exactly `requested`, or more.

e.g. in the example above, the Cold storage implementation would return all 1500 events, despite `requested == false`. It would be equally valid to first return 1000 and then return 500, or even to return 1200 and then 300.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+1. Builds can generate an enormous amount of build events - storing everything
+   in the same database as all of the other core-Concourse information means
+   that, over-time, the `build_events` tables will tend to dominate (in many
+   cases). For instance, in our (fairly small) Concourse deployment at
+   `ci.concourse-ci.org`, build events comprise ~70% of the total ~13GB used.
+   While storage is cheap, I can imagine this could result in general sluggishness
+   in larger Concourse deployments as indexes get large. While much of this can be
+   alleviated by more aggressive build log reaping, that can introduce performance
+   problems of its own, since deleting lots of build events can be taxing on
+   Postgres. I'm sure it's fairly obvious at this point that I really don't know
```
jchesterpivotal, line 22: Both the performance overhead and deletion cost would be improved by using table partitions.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+   much about running Postgres/Concourse at scale, so let me know if what I'm
+   saying here doesn't hold true.
+1. Postgres wasn't designed for handling huge amounts of text data. Again, I
+   don't have much evidence/expertise to back this up, but having many large
+   rows in a `build_events` table strikes me as sub-optimal (compared to a system
+   designed for handling large amounts of text).
```
jchesterpivotal, line 28: Correct, PostgreSQL's text handling is somewhat suboptimal. But it's not _terrible_. Concourse is using it mostly for long-term storage of logs -- it's not being used as a fulltext search engine. In terms of storage, PostgreSQL automatically compresses `text` fields _if_ they are >2k long. That won't work for most events rows unfortunately.

What would make the biggest difference here would be a columnar store for the events table, since compression along a column of build logs is likely to be _very_ compressible indeed. The relevant 'zedstore' technology is being worked on but is a ways off, so irrelevant for this RFC.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+1. Storing build events in a system designed for handling text data (like
+   Elasticsearch, for instance) allows you to do some pretty cool querying that
+   isn't feasible in Postgres.
+1. Gaining observability into build events isn't great. The syslog drainer
```
jchesterpivotal, line 32: Agreed, I've had to do quite a lot of work when I've wanted to extract and transform this data for analysis.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+
+
+# New Implications
+
+* By default, there will be no change - Concourse will default to using a
+  Postgres `EventStore` using the same main Postgres database.
+* Operators will be able to configure a different store
+* If they configure a new store, they won't be able to access any existing
+  build events that are currently stored in Postgres without migrating them
+  over (how should we recommend doing this?)
```
jchesterpivotal, line 457: I think it should be a standalone tool, not part of the ATC. Otherwise you're forcing folks to upgrade to get it, or to get bugfixes to such a migration tool.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+type EventNotificationBus interface {
+    Notify(build Build) error
+    Listen(build Build) (<-chan struct{}, error)
+}
+```
+
+But I'm not sure if there's much value in that.
+
+
+## <a name="migrations">Migrations</a>
```
jchesterpivotal, line 394: You don't need to provide anchors if publishing on github, they add them automatically.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+
+Currently, we have migrations to set up the `build_events` table, as well as
+`TRIGGER`s for creating/deleting `pipeline_build_events_x` and
+`team_build_events_x` tables when pipelines and teams are created/deleted.
```
jchesterpivotal, line 398: Aside: I wish this were achieved with partitions rather than templated tables.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,457 @@
...
+  internally, but I guess we may want to keep it for backwards compatibility?
+  It's not exposed by `fly` or `go-concourse`, either.
+
+* Would any event stores benefit from bringing their own notification bus
+  implementation, rather than relying on Postgres' notification bus? Does using
+  the Postgres notification bus defeat the purpose of having an external store in
+  any way (e.g. could this be a bottleneck)?
+
+* Users won't be able to access build events that exist in the main Postgres
+  database if they switch to a new backend store. How can we make the
```
jchesterpivotal, line 440: I think there are two use cases being mixed here.

The first is "tell me what just broke", the thing Concourse UX is optimised for. That data works best in PostgreSQL, close to the ATC.

The second is "I need a complete archive of everything ever". Concourse enables that but isn't really putting it front and centre. The overhead has been historically bad enough that the archival need was at cross-purposes with the diagnostic goal. This is where draining to secondary stores is useful. 

If you have a secondary store _in addition_ to the main store, then setting short build histories on the _main_ store, but not the _secondary_ store, is a relatively safe compromise for most installations.

053-configurable-event-store/proposal.md
cirocosta: I thought that was a fairly new feature, but it seems like it isn't:

Screen Shot 2020-05-12 at 2 09 31 PM

that's nice (saying that especially because https://github.com/concourse/concourse/issues/5391#issuecomment-607346689)

053-configurable-event-store/proposal.md
cirocosta: @pivotal-jwinters @vito (I guess?) is this something considered back then?

053-configurable-event-store/proposal.md
jchesterpivotal: Yes, though the advances in table partition capabilities and performance between each version are substantial.

(Also the Concourse postgresql support policy should change, but that's another discussion).

jchesterpivotal: In particular:

1. Foreign key and uniqueness constraints were not enforced across partitions until v11.
1. Multiple partition keys were not supported until I think 11 or 12. These let you partition the main multiple ways if it proves to be useful. Partitioning by team and date seems reasonable, but studying existing production databases would be better.

053-configurable-event-store/proposal.md
aoldershaw: Ah, good to know!

053-configurable-event-store/proposal.md
aoldershaw: Yeah, that assessment makes sense. I think the proposed `EventStore` interface should be able to support this (similar to the "Cold Storage" example I provided, but could drain to an arbitrary secondary `EventStore`). Pair that with build log reaping to clear out old Postgres logs, and we can keep recent builds close to the ATC in Postgres while archiving everything in a secondary store (which could be S3, Elasticsearch, etc)

053-configurable-event-store/proposal.md
aoldershaw: in the Partitioning doc for 9.5/9.6, it seems to be what we're doing (creating a table for each pipeline/team that `INHERITS` from the `build_events` table)

@jchesterpivotal could you elaborate on what you mean by table partitions? I notice in Postgres 10+ there's "declarative partitioning" (e.g. `CREATE TABLE ... PARTITION BY ...`) - is that what you're referring to?

jchesterpivotal: "Table partitioning" is the general name for the feature set of "some way of physically subdividing tables while giving some seamless single logical view". You can do it with table inheritance but it wasn't really designed for that purpose. The later features are more "real".

Especially because declarative partitioning is handled explicitly by the database, which can continuously find new performance gains in each release. With inheritance+triggers insert performance will be bound by your own code. And similarly at query time: a declarative partitioning scheme gives the query planner reliable information about which partitions can be ignored in a query plan.

aoldershaw: Though, I guess with our current Postgres support policy, we'll be stuck with table inheritance until 9.6 is no longer supported (not until late 2021) 🤷

053-configurable-event-store/proposal.md
deepakmohanakrishnan1984: I really like the option 2 having FallbackEventStore composing main and old eventstore. To add an extension to it, the fallback eventstore can also perform the migration(at background) when it read from an old eventstore.

```go
func (f *FallbackEventStore) Get(ctx context.Context, build db.Build, requested int, cursor *Key) ([]event.Envelope, error) {
   
  // first, try the main event store:
  ...

   events, err = f.Fallback.Get(build, requested, cursor)

   if err == nil && len(events) > 0 {
        //perhaps run this in the background to migrate this events to main eventstore
        f.EventStore.Put(ctx, build, events)

        return events, nil
    }
    return nil, err
}
```

The build event logs will be migrated only when a user access it for the first time. In most cases, users might not be interested looking at historical build events, so the expensive migration of all historical build event logs to main eventstore can be skipped and will be reaped after its retention period.

deepakmohanakrishnan1984: I really like the option 2 having FallbackEventStore composing main and old eventstore. To add an extension to it, say lazy migration 😄 , the fallback eventstore can perform the migration(at background) when it read the events from an old eventstore.

```go
func (f *FallbackEventStore) Get(ctx context.Context, build db.Build, requested int, cursor *Key) ([]event.Envelope, error) {
   
  // first, try the main event store:
  ...

   events, err = f.Fallback.Get(build, requested, cursor)

   if err == nil && len(events) > 0 {
        //perhaps run this in the background to migrate this events to main eventstore
        f.EventStore.Put(ctx, build, events)

        return events, nil
    }
    return nil, err
}
```

The build event logs will be migrated only when a user access it for the first time. In most cases, users might not be interested looking at historical build events, so the expensive migration of all historical build event logs to main eventstore can be skipped and will be reaped after its retention period.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,465 @@
...
+  choose a different type of `Key`. Regardless of the type of `Key`, if the goal
+  is to fetch all events, `cursor` should initially point to `nil` (but the
+  pointer itself must not be `nil`). For instance, you can iterate over all
+  events for a completed build like so:
+
+  ```go
+  var cursor Key
+  batchSize := 1000
+  for {
+      events, err := eventStore.Get(build, batchSize, &cursor)
```
deepakmohanakrishnan1984, line 133: minor nitpick: I noticed the `ctx context.Context` is added to the interface functions recently. It is missed in the param list here and other embedded code snippets.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,452 @@
...
+haven't changed `build_events` since the initial migration).
+
+
+## Configuration
+
+One thing I haven't really explored is how operators would configure the
+backend store. I think it could be similar to how Credential managers are
+configured - there are groups of independent configuration options for each
+possible `EventStore`. If certain flags are configured for a given backend, use
+that backend.
```
ari-becker, line 419: I think that it would be desirable to design build event stores in such a way that allows them to be built and released independently from the rest of Concourse, similar to how Terraform eventually spun providers out of its main source tree so that they could be built and released independently, typically written and maintained by the commercial entities backing the providers.

For example, a user might wish to export build logs to Coralogix, which is a log management and analysis platform. As Coralogix is a commercial platform, I would fully expect that a Coralogix build event store plugin would be written and maintained by Coralogix and not by the Concourse team / Pivotal.

(full disclosure: I work for Coralogix and we would probably (but I'm intentionally being non-committal here) be interested in building out a plugin for our own platform as part of dogfooding our own product)

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,452 @@
...
+    EventStore
+    ColdStorage ColdStorage
+}
+
+func (c *ColdStorageEventStore) Finalize(ctx context.Context, build Build) error {
+    if err := c.EventStore.Finalize(ctx, build); err != nil {
+        return err
+    }
+    // perhaps run this in the background
+    return c.migrateToColdStorage(build)
```
ari-becker, line 253: I think this comment belies a lot of underlying complexity, where this design should go towards a plugin model per-provider, like Terraform (see comment below about adding providers), which uses an RPC architecture. See for example: https://www.terraform.io/docs/extend/how-terraform-works.html

This kind of multiple-binary architecture is necessary for both a) proper queuing and background processing of events as they are sent to external providers, and b) allowing additional providers to be written and developed independently of the upstream project.

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,452 @@
...
+
+func (s *SyslogForwardingEventStore) Put(ctx context.Context, build db.Build, events []atc.Event) error {
+    if err := s.syslogForwardEvents(build, events); err != nil {
+        return err
+    }
+    return s.EventStore.Put(ctx, build, event)
+}
+```
+
+How about migrating build events to cold-storage after they're completed:
```
ari-becker, line 235: I wonder if the desire to design for cold storage doesn't mean that `EventStore.Get` should be some kind of optional operation. Usually the trade-off with cold storage is that it reduces storage costs while increasing access costs, and we wouldn't want to subject users to a surprisingly high bill just because they accessed some part of Concourse in a way which invoked the `EventStore.Get` on cold-storage.

Maybe a better design would be one where `EventStore.Get` can return a link to where the build events can be found? Then the user can more intentionally make the decision to access cold storage and invoke those costs.

053-configurable-event-store/proposal.md
aoldershaw: That's very interesting - I hadn't put much thought into an architecture like that, mostly because the precedent seems to be package and release everything from within Concourse (at least in terms of credential managers and metrics emitters). I think it'd be cool to for build event stores to be written as plugins external to the Concourse binary, and it'd be really cool to have experts in the domain be the ones to implement/maintain it.

I guess one concern is the added operational complexity, though I imagine it's not too bad. Like Terraform, we could package some plugins directly into Concourse (it could just be the Postgres plugin so operators don't need to bring their own plugin for the default behaviour).

053-configurable-event-store/proposal.md
aoldershaw: Yeah, the comment was a bit of a half-baked idea. From what I can tell, any sort of background processing we have now is just done through separate goroutines, everything residing in the same process. For instance, metric emitters run on an [`emitLoop`](https://github.com/concourse/concourse/blob/36cbf5ecf628dba92b6b4c08cfa536820638f313/atc/metric/emit.go#L136-L140) that reads from a Go `chan` that receives messages: https://github.com/concourse/concourse/blob/36cbf5ecf628dba92b6b4c08cfa536820638f313/atc/metric/emit.go#L129-L133

All that to say, we *could* (and currently do) perform some sort of background processing together without the use of plugins running in their own process and communicating via RPC. That's definitely not to say we *should* do things this way (though, admittedly, I don't fully grasp all of the implications of taking one approach over the other).

053-configurable-event-store/proposal.md
aoldershaw: Hmm good point. I'm hesitant to modify the `EventStore` interface to explicitly make `Get` optional, since that seems pretty specific to this use case (and I think it generally makes sense for there to be `Get` functionality).

I guess this is also modifying the API (albeit in a different way), but we could introduce a special build event type that includes the link to cold-storage (e.g. `event.InColdStorage`), and a ColdStorage implementation would return that single build event when called. The event consumers (the UI and Fly) could know to interpret that event type specially, perhaps prompting the user to confirm they want to fetch from cold storage. This prompt could also be bypassed with a special `Key` (passed as the `Last-Event-ID` to the build events API endpoint) to indicate "yes, I know I'm fetching from cold storage, and I'm okay with that".

aoldershaw:
Thanks for your feedback @agurney! You bring up some really good points. Going to preface my response by saying I'm not an expert on this stuff, so please bear with me if I say anything that doesn't make sense 😄- if you disagree with anything, please let me know!

> I'd like to probe a bit on what's envisaged here, because I am uneasy about bringing a lot of fresh complexity into Concourse itself as far as log forwarding. There are existing tools like logstash, fluentd, or NiFi, which can do all sorts of things in terms of fanout, enrichment/transformation, and high-performance streaming

I looked briefly into logstash during my investigation into Elasticsearch, but I've never used any log forwarding tools myself so don't fully grasp the value-add for Concourse build events (besides performance, as I imagine they're quite well optimized). To me, it seems like one benefit is turning raw unstructured logs into structured data (I suppose that's the enrichment/transformation you bring up), but Concourse build events already are structured to a certain degree. Another benefit is the decoupling of inputs from the outputs - as you mention, Concourse could just emit to syslog, and logstash/fluentd/NiFi could forward the logs to the correct destination. However, given the fact that Concourse will need to retrieve logs from these destinations as well, there's an implicit coupling that still needs to exist somewhere:

> Then the implementation is that the builds DB table grows an extra column for the remote storage URL (maybe with other data as needed for retrieval, such as auth tokens or whatever). When build logs are requested, and not found locally, we fetch them via that URL instead.

I'm not quite sure what you mean by "storage URL" - do you mean something like a `https://...` API endpoint that Concourse can just call out to without needing to know what's on the other side, or something with a provider-specific scheme that Concourse has to interpret: e.g. `elasticsearch://{elasticsearch-url}/builds/{build_id}`.

If you mean the former, I'm not convinced this will work in general. For instance, if we emit build events to Elasticsearch, we could store the `/{index}/_search` API endpoint with an appropriate query, but how would we go about e.g. pagination, or even parsing the results into a common form?

If you mean the latter, wouldn't it be easier to just have an implementation of an `EventStore` that has a `Get` method (that can live outside of the core-Concourse codebase)?

Let me know if I completely misunderstood what you meant here 😄

> So I think the remaining non-existing piece is retrieving logs from their remote home, for display through Concourse. I'm imagining a world where logs always go to the local Postgres for at least a few days, but some operators will configure aggressive local reaping combined with forwarding into another store

This sounds a lot like what @jchesterpivotal was saying here: https://github.com/concourse/rfcs/pull/53#discussion_r424025339 (keep recent builds as close as possible to the ATC, and optionally configure an external event store for archiving builds). I think that's a good way to look at the problem. I did reply to his comment mentioning that I think this could be modelled with the proposed `EventStore` interface, but it's probably worth revisiting the proposed design with that in mind.

> I'm not sure that it should be Concourse's job to decide when remote logs should be reaped. Once we've passed the data on, it's the other system's job to implement its own retention policy, according to whatever criteria happen to make sense... We may also not want to force log deletion just because a pipeline has been deleted - maybe those logs are still useful for analytics

That's a really good point, I hadn't considered that! Do you think there's value in Concourse letting the external system know "by the way, this pipeline no longer exists, so you may want to delete the build events", and the system can choose to delete the events or not?

> How should Concourse be expected to authenticate itself to a remote log store, given that not all users should see all logs, and remote systems have their own idea of how access control works? (That's still the case for the current proposal as well, since a given `EventStore` would have to implement that logic in some fashion.)

I'm not 100% sure of what types of access control remote systems have, so maybe you could speak more to that. I kind of envisioned it as working like it does currently with Postgres: the ATC is provided credentials that would grant access to all build events, and Concourse implements handles user access controls by team at the API. So, `EventStore` implementations assume they're able to access build events for any build with the credentials they're configured with, and it's up to Concourse not to serve build events to users who aren't able to access them.

053-configurable-event-store/proposal.md
aoldershaw: I was thinking about this a bit more. For *some* use cases of the proposed `EventStore`, having their implementations be invoked directly (as opposed to over something like gRPC) is preferable. For instance, invoking the Postgres `EventStore` in memory allows us to perform queries database objects (e.g. [also deleting all of a team's pipelines when deleting a team](https://github.com/concourse/concourse/pull/5651/commits/18832d57e1a175b6931ef7a845c01a87c11836e8#diff-35004ec688e559fe368187a2c0638fefR266)). For certain applications of `EventStore` composition (e.g. [for secret redaction](https://github.com/aoldershaw/rfcs/blob/config-event-store/053-configurable-event-store/proposal.md#eventstore-composition)), handling the operations in memory is beneficial.

So, what if `EventStore`s are just an in-memory type, but there's a `PluginEventStore` implementation that functions as a gRPC client? Which plugin (gRPC server) to use could be configured via flags to Concourse, and any environment variables such as `CONCOURSE_EVENTSTORE_PLUGIN_*` could be passed along to the plugin on startup to configure e.g. authentication (similar to how we support [custom configuration for Guardian](https://github.com/concourse/concourse/blob/72c7cdad6f2ef05036c6c9fb44a2fc5eee6cbb88/worker/workercmd/guardian.go#L116)).

What do you think of that approach?

053-configurable-event-store/proposal.md
```diff
@@ -0,0 +1,452 @@
...
+One thing I haven't really explored is how operators would configure the
+backend store. I think it could be similar to how Credential managers are
+configured - there are groups of independent configuration options for each
+possible `EventStore`. If certain flags are configured for a given backend, use
+that backend.
+
+
+# Open Questions
+
+* Does `Key` need to be `interface{...}`, or can it be more specific like `uint`?
```
aoldershaw, line 424: In my [experimental PR](https://github.com/concourse/concourse/pull/5651) that introduced an Elasticsearch implementation of an `EventStore`, I found that I needed a composite key of `(timestamp, tiebreaker)` (where both `timestamp` and `tiebreaker` are `int64`s) to uniquely identify events, so a single `uint` isn't good enough (without relying on Postgres to generate auto-incrementing IDs).

I was experimenting with plugin-based `EventStore` implementations at the suggestion of @ari-becker, and realized that if the plugin runs in a separate process (i.e. we use something other than [Go plugins](https://golang.org/pkg/plugin/), which seem very limiting), we'll need some way of sending the `Key`s over the wire while retaining their ordering information. This isn't so easy with an arbitrary `Key` type, so I got back to thinking about restricting the `Key` type to something more concrete that has an implicit ordering.

From the Postgres and Elasticsearch implementations, I got the idea of using a `[]int64`. Using `timestamp` and `tiebreaker` could be used for many backends (the `timestamp` comes from the event, and `tiebreaker` is an in-memory counter - more details in the PR) - but for backends that support a more native keying approach (like Postgres' auto incrementing `event_id` or line numbers in file-based storage like S3 or local FS), we could make use of that instead. I wonder if there are backends that won't fit into the `[]int64` mold - if anyone has a system in mind, I'd love to hear about it. I also wonder if the `Key` could always be a `[2]int64`, or if there's value in having an arbitrary length slice.

Another approach is to avoid the `Key`/cursor approach altogether, which I'd definitely be open to ideas on.

jchesterpivotal: There's basically no way to establish sane global ordering without the whole nightmare of distributed consensus. I heartily discourage trying. Leave the business of sorting out ordering to consumers, who can choose what interpretations to apply.

But if what you need is _uniqueness_, that's much more straightforward. You have broadly two options:

1. UUIDs, specifically v4. Basically a 122-bit random number, so they can be generated without coordination. They are widely supported data type. PostgreSQL has a native type and the uuid-ossp module can generate them in the database very fast. They're supported in pretty much everything everywhere.
2. Natural keys. That is, you determine what the distinguishing fields are and then you digest them. This is isomorphic to the database normalisation exercise of identifying the natural key for a table: what is the field or combination of fields which absolutely defines uniqueness? In this case it might be something as simple as `(something_identifying_the_database, event_id)`. As long as the database is unique and the event_id is unique within the database, it "should" be unique overall.

I prefer option 1 to option 2. It's easier to implement and more robust to oversights. Natural keys are easy to describe but once you pick one, it's hard to fix if you discover you made a mistake and have collisions. Plus people are liable to start trying to rely on the natural key components, meaning you wouldn't be able to change it.

tl;dr UUIDs errywhere

aoldershaw: I think that's good advice - in my experimental PR for Elasticsearch, I made the naïve assumption that I could trust the timestamp on build events for ordering, when obviously isn't the case. Builds are tracked by a single ATC at a time, so at least time-based inconsistencies aren't as prevalent, but in general, it's a hacky "solution" (that doesn't really solve the problem). So, I agree that we shouldn't try to achieve distributed consensus.

> Leave the business of sorting out ordering to consumers, who can choose what interpretations to apply

This makes sense, but does this mean that Concourse will have to fetch all events for a build  (and sort them in memory) before it can serve any events? Or am I misunderstanding what you mean?

Another option that I suggested as a possibility is to use natural keys of `(build_id, event_id)`. For backends that don't support auto incrementing (e.g. Elasticsearch), rely on Postgres sequences to generate these `event_id`s. This way, we can take advantage of the indexing options of tools like Elasticsearch to perform sorting/pagination outside of Concourse.

jchesterpivotal: > This makes sense, but does this mean that Concourse will have to fetch all events for a build (and sort them in memory) before it can serve any events? Or am I misunderstanding what you mean?

Sorting in the database is fine, using the internal sequential IDs (ie the primary key). Those are typically monotonic, so they establish a local order. But they're only an ordinal scale (this-before-that), not an interval score (this-10-seconds-before-that) or ratio score (this-10-percent-longer-than-that).

053-configurable-event-store/proposal.md
evanchaoli: Maybe just not migrate. Instead, add a flag to table `build` to indicate where is build logs of the build, PG or ES. So that reads old build logs from PG and new build logs from ES.

clarafu: Just a suggestion for another way to do the migration, I wonder if there could be a way to migrate over all the build events to their new event store without stopping the database? This will make it so that operators wanting to move over to a new build event store does not need to wait for the migrations to finish before starting concourse.

I imagine something like the operator would continue to have a running concourse deployment and the build events would currently be fetched from the postgres database. While the deployment is running, some tool can be used to migrate all the build events within the postgres database (old build event store) into the new build event store. Then once it has finished, the operator can redeploy concourse and point it at the new build event store. There will need to be some method for discarding the old data (maybe concourse can do that at startup if you configure a new event store? or maybe it can be part of the tool to wipe out build events?) 

I wonder for operators that want to configure a new build event store, they might want to benefit from being able to query or search for things within the build events. If we were to do the method of migrating on the fly, that would mean only certain build events would be migrated over. So taking that into account and the fact that operators that configure a new build event store most likely would like all their builds to be in this new store, I might be inclined to thinking that migrating all the data together would be preferable? I could totally be wrong about this though.

aoldershaw: @clarafu that definitely makes sense - having a separate tool that just points to the running Postgres is very much in line with https://github.com/concourse/rfcs/pull/53#discussion_r424022341.

FWIW the idea behind the "on the fly" approach is that it'd eventually migrate all build events over (not just for new events) - it'd essentially be doing what the tool you're suggesting is doing, but it's started by the ATC. I think it's beneficial to keep it separate though, for the reasons mentioned by @jchesterpivotal.

clarafu: Ohh I see! Sorry I missed where you clarified it by saying it will `build reaper that migrates old builds to the new store`. And yeah that is a good call that allowing the operator to migrate before upgrading is definitely a good idea.

053-configurable-event-store/proposal.md
evanchaoli: I doubt in-memory EventStore, because that implies every ATC will duplicated store build logs, which might not fit to huge clusters where hundreds pipelines generate huge amount of build logs. But I think plugin is a good idea, at least, a PG plugin would allow us to store build logs to a separate database, if the plugin is well designed, it may even support to dispatch build logs to multiple databases, which would definitely benefit scalability of Concourse. Also, users will have the flexibility to store build logs in various backends, ES, and whatever.

taylorsilva:
Not doing this. General lack of interest from others too from what I'm seeing.

PR #59, opened 2020-06-05
RFC: Propose static configuration for Concourse instances

ari-becker:
[Rendered](https://github.com/ari-becker/rfcs/blob/statically-configured-atc/059-statically-configured-atc/proposal.md)


059-statically-configured-atc/proposal.md
```diff
@@ -0,0 +1,192 @@
...
+
+This RFC proposes replacing this call structure by adding the following section
+to the configuration YAML:
+
+```yaml
+options:
+  # ... omitted for the sake of brevity ...
+teams:
+  # ... omitted for the sake of brevity ...
+projects:
```
ari-becker, line 116: it may be better to nest `.projects` under `.teams`. What do you think @vito ?

059-statically-configured-atc/proposal.md
```diff
@@ -0,0 +1,192 @@
...
+  is being driven by engineering leadership (i.e. top-bottom adoption), then
+  leadership is challenged to orchestrate the synchronization of configuration
+  across a wide variety of Concourse instances that have been set up for
+  engineering teams.
+* For any sufficiently long-lived Concourse instance, database usage grows over
+  time, and the underlying PostgreSQL database grows vertically to match,
+  increasing the costs for end teams. Reducing reliance upon the underlying
+  PostgreSQL database has a direct, positive effect on instance performance and
+  costs incurred by Concourse administrators.
+* Perusing the current configuration for a Concourse instance, when so much of
```
clarafu, line 24: I'm not exactly sure what this point means, is this talking about the `web` configuration of the postgres instance (https://github.com/concourse/concourse/blob/master/atc/atccmd/command.go#L130) or the configuration of postgres itself (which is done depending how you deployed it through helm, bosh, etc).

059-statically-configured-atc/proposal.md
```diff
@@ -0,0 +1,192 @@
...
+options:
+  external_url: https://concourse.example.com
+  enable_lets_encrypt: true
+  postgres_user: concourse
+  postgres_password: ((CONCOURSE_POSTGRES_PASSWORD))
+```
+
+## Replacing `fly set-team`
+
+Doing so permits the configuration file to include additional configuration that
```
clarafu, line 63: This is definitely an interesting idea, do you have a use case where an operator would like to restrict only being able to modify teams through the configuration file? My only concern with this is that it would require the operator to have to restart the web nodes everytime they want to make a change to the teams, but I do understand how it would be nice to have the true list of teams in the configuration file for the cluster. I guess I'm more curious of the "why" this would be nice for operators, to be able to be strict with the teams configuration?

059-statically-configured-atc/proposal.md
```diff
@@ -0,0 +1,192 @@
...
+Permit running a `web` instance with a
+`--static-instance-configuration=/path/to/config.yaml` flag. If the
+`--static-instance-configuration` flag is provided, then no other command-line
+flags will be permitted; attempting to provide other flags (excepting the
+`--validate` flag, see below) will result in `concourse web` exiting with an
+error. Sending `SIGHUP` to the `web` process will ask the `web` process to
+reload this file, without restarting the process.
+
+The contents of this configuration YAML file will permit listing any of the
+existing options under `.options`. Sensitive strings may be substituted with
```
clarafu, line 48: Is there a reason why the configuration fields needs to be under `options:`? In my proposal for starting concourse with a configuration file, I had it so that you can use one configuration file to start up your web and worker nodes. So the web configuration would be under `web:` and the worker configuration would be under `worker:`.

clarafu (commented):
Hi! I was actually already working on an approach towards starting concourse using a config file, the issue is [here](https://github.com/concourse/concourse/issues/6434) and I have an open draft PR [here](https://github.com/concourse/concourse/pull/6893). I'd love to get your feedback on any of that if you would like!

ari-becker:
I'm cleaning up my old PRs, I left the position where I was using Concourse professionally.
I'd be happy to re-open if someone considers this to still be relevant.

PR #62, opened 2020-06-23
WIP: worker pool

evanchaoli:

### DIFF
--- a/062-worker-pool/proposal.md
+++ b/062-worker-pool/proposal.md
@@ -1,0 +1,157 @@
+# Summary
+
+This proposal outlines the support for worker pool. A worker pool is a group
+of Concourse workers, that can be associated with one or multiple teams.
+
+# Motivation
+
+One or multiple Concourse workers can be associated with a team by adding 
+`--team <team name>` option when launching `concourse worker`. Then all 
+pipeline containers of the team will be dispatched to the workers. The 
+relationship between workers and teams are N-to-1:
+
+```
+  Workers (0..n) <----------------> (0..1) Team
+```
+
+This is a useful mechanism to isolate some teams' workload or to provide
+dedicate workers to some teams.
+
+In practice, some big projects may split their pipelines to multiple small
+teams. When a project that has multiple teams requests dedicate workers,
+it is hard to give. Because the project actually wants some workers sharing
+with all of its teams, but Concourse only support workers per team.
+
+For example, a big project of my company has 100 teams. By evaluating
+workload of those teams, the project requests 20 dedicate workers. However, due
+to the restriction of 1 worker serving at most 1 team, we have to deploy at
+least 100 workers (1 worker for a team) to the project. Moreover, each team 
+having a single worker is unreliable, if the worker is down, then all builds of
+the team will fail. If we deploy 2 workers for each team for redundancy, then
+we need to deploy 200 workers. Comparing 200 workers we need to deploy with 20
+workers they request, it is a huge waste.
+
+# Proposal
+
+With introducing a new concept "worker pool", we build N-N relationship between
+workers and teams:
+
+```
+  Workers (0..n) <----> (1) Worker Pool (1) <------> (0..n) Teams
+```
+
+* A worker pool may contain 0 or multiple workers.
+* A worker can be added to at most 1 worker pool.
+* A worker may also not belong to any worker pool.
+* If a worker is added to a worker pool, then it cannot be associated with a
+team directly.
+* If a worker is not added to any worker pool, then it can still be associated
+with a team directly for backward-compatibility.
+* A worker pool may be assign to 0 or multiple teams.
+* A team can be associated with at most 1 worker pool.
+* A team can still be associated with a worker directly for backward-compatibility.
+* A team may not associate with any worker pool or team for backward-compatibility.
+
+The logic of choosing a worker for a step will look like:
+
+```go
+   team := step.TeamName()
+   var candidateWorkers = []Workers
+   if team.WorkerPool() != nil {
+      candidateWorkers = team.WorkerPool().Workers()
+   } else if team.Workers() != nil {
+      candidateWorkers = team.Workers()
+   } else {
+      candidateWorkers = AllNoTeamAndNoPoolWorkers()
+   }
+   chosenWorker, err := chooseWorker(candidateWorkers, step.Tags)
+```
+
+## Tag
+
+There is no impact to tags.
+
+When launching `concourse worker`, `--tag` option can still be used to mark the worker
+with some tags. As the above code demonstrates, worker pool only impacts candidate workers,
+but the algorithm of choose a worker from candidate workers.
+
+## fly
+
+### To list all worker pools:
+
+```
+$ fly -t <target> worker-pools
+```
+
+### To create a worker pool:
+
+```
+$ fly -t <target> set-worker-pool --create --name <pool name> [--public-worker-key=<public ssh key file>]
+```
+
+When creating a worker pool, it allows to optionally specify a ssh public key 
+file. If a public key is configured with a pool, then workers trying to join
+the pool must use the corresponding private key.
+
+### To delete a worker pool:
+
+```
+$ fly -t <target> delete-worker-pool --name <pool name>
+```
+
+A worker pool should only be able to be deleted if there is no team associated 
+with it, and no worker in the pool. 
+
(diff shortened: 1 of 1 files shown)


062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
...
+
+* A worker pool may contain 0 or multiple workers.
+* A worker can be added to at most 1 worker pool.
+* A worker may also not belong to any worker pool.
+* If a worker is added to a worker pool, then it cannot be associated with a
+team directly.
+* If a worker is not added to any worker pool, then it can still be associated
+with a team directly for backward-compatibility.
+* A worker pool may be assign to 0 or multiple teams.
+* A team can be associated with at most 1 worker pool.
```
vito, line 51: Interesting - do you think this is a necessary restriction, or is this just being conservative to start out?

It kind of seems like a team could be associated with multiple, and just union all the worker pools together to get one big set of candidates. The final selection would then be based on platform/tag configuration within the set of workers.

For example, maybe there's a pool of Windows workers and a pool of Darwin workers that teams can opt in to. I guess those could be in one big pool, but maybe there'd be some reason to make those platforms opt-in - maybe the operators are  being careful about sharing them because they don't containerize as effectively.

🤷‍♂️

062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
...
+* A team can still be associated with a worker directly for backward-compatibility.
+* A team may not associate with any worker pool or team for backward-compatibility.
```
vito, line 53: not sure what this means - is there a typo? 🤔

062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
...
+* A worker pool may contain 0 or multiple workers.
+* A worker can be added to at most 1 worker pool.
+* A worker may also not belong to any worker pool.
+* If a worker is added to a worker pool, then it cannot be associated with a
+team directly.
+* If a worker is not added to any worker pool, then it can still be associated
+with a team directly for backward-compatibility.
+* A worker pool may be assign to 0 or multiple teams.
+* A team can be associated with at most 1 worker pool.
+* A team can still be associated with a worker directly for backward-compatibility.
```
vito, line 52: Is this the same as this line from above?

> If a worker is not added to any worker pool, then it can still be associated with a team directly for backward-compatibility.

If so, I think this can be dropped; I find the other description clearer, since it reflects the direction of the relationship (many workers -> one team, whereas this sounds like one team -> one worker).

062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
...
+The logic of choosing a worker for a step will look like:
+
+```go
+   team := step.TeamName()
+   var candidateWorkers = []Workers
+   if team.WorkerPool() != nil {
+      candidateWorkers = team.WorkerPool().Workers()
+   } else if team.Workers() != nil {
+      candidateWorkers = team.Workers()
+   } else {
```
vito, line 64: Having this be either-of seems kind of restricting - for example, what if a team wants to use the shared pool, but also has their own self-maintained worker that they would like to register? For example, sometimes teams will use a shared pool of workers but also run their own Mac Mini for running Darwin builds for iOS development.

I guess we could represent that as its own pool, but that would also imply allowing a team to have multiple pools.

062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
...
+### To list all worker pools:
+
+```
+$ fly -t <target> worker-pools
+```
+
+### To create a worker pool:
+
+```
+$ fly -t <target> set-worker-pool --create --name <pool name> [--public-worker-key=<public ssh key file>]
```
vito, line 89: Is the `--create` flag necessary?

062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+When creating a worker pool, it allows to optionally specify a ssh public key 
+file. If a public key is configured with a pool, then workers trying to join
+the pool must use the corresponding private key.
+
+### To delete a worker pool:
+
+```
+$ fly -t <target> delete-worker-pool --name <pool name>
```
vito, line 99: For consistency with other commands:

```suggestion
$ fly -t <target> destroy-worker-pool --name <pool name>
```

062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+A worker pool should only be able to be deleted if there is no team associated 
+with it, and no worker in the pool. 
+
+## To associate a team with a worker pool
+
+```
+$ fly -t <target> set-worker-pool --assoc --name <pool-name> --team <team name> [--team <team name2> ...]
```
vito, line 108: Ah, this must be why `--create` was necessary. This feels like it should be a separate command instead - maybe `fly attach-worker-pool`?

062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+When associating a team with a worker pool, it should verify that the team
+currently does not associate with any other worker pool.
+
+## To dissociate a team from a worker pool
+
+```
+$ fly -t <target> set-worker-pool --dissoc --name <pool-name> --team <team name> [--team <team name2> ...]
```
vito, line 117: Maybe `fly detach-worker-pool`?

062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
...
+`--team <team name>` option when launching `concourse worker`. Then all 
+pipeline containers of the team will be dispatched to the workers. The 
+relationship between workers and teams are N-to-1:
+
+```
+  Workers (0..n) <----------------> (0..1) Team
+```
+
+This is a useful mechanism to isolate some teams' workload or to provide
+dedicate workers to some teams.
```
ari-becker, line 18: Maybe it's just that I run a (I guess, relatively) small Concourse instance, but I'm struggling to understand the underlying motivation here.

A core tenet of Concourse is avoiding worker state, with tags being a compromise to allow for e.g. hardened network architectures. Why do teams care about isolating their workload in the first place? How does one team's workload affect another team's workload? If it's to ensure worker availability for their projects, doesn't it make more sense to work on a feature like auto-scaling workers (i.e. "grow the pie") rather than a feature to protect and hoard "precious" workers (i.e. "divide the pie")?

If an organization *really* wants to provide a high level of isolation, why not set up isolated Concourse instances? Why only isolated Concourse worker pools?

062-worker-pool/proposal.md
```diff
@@ -0,0 +1,157 @@
+# Summary
+
+This proposal outlines the support for worker pool. A worker pool is a group
+of Concourse workers, that can be associated with one or multiple teams.
```
ari-becker, line 4: For what it's worth, when I originally heard of "worker pools", I was thinking more along the lines of allowing different Concourse installations to coordinate with each other to access a common pool of workers, which would make it much easier to launch new Concourse installations for different teams by giving them access to some workers already by default and not have to set up workers separately in order to get started.

062-worker-pool/proposal.md
evanchaoli: This is a good point. I agree that a team may have multiple pools.

062-worker-pool/proposal.md
evanchaoli: I guess this is an incomplete statement. I couldn't understand it by myself now. Just deleted it.

062-worker-pool/proposal.md
evanchaoli: My original thoughts is, a worker pool is nothing more than a name, thus workers with different platforms, tags can join the same pool. Given a pool may attach to multiple teams, if a team can also attach to multiple pools, then the relationship between pool and team is N-to-N, which will make management more complicated.

But I can also imagine an use case like, teamA and teamB share a Darwin pool as Darwin workloads are limited, but each of them has a dedicate Linux pool.

evanchaoli:
@vito Thanks for your early review that makes my work easier.

taylorsilva:
@muntac

taylorsilva:
Closing this RFC. I think the idea is cool and useful for any org that has a large Concourse cluster with a lot of teams. The lack of interest in this idea though tells me that this group is either very small or just not interested in this feature.

PR #73, opened 2020-09-01
implement mTLS

nickhyoti:
Proposal to implment mTLS authentication within Concourse Web


nickhyoti:
Is anyone able to help with this?  It seems to be stuck?

avanier:
This is quite similar to the work I did on concourse/concourse#6017.

It generally seems on the up-and-up to me, but usually I go by the checks to make sure I'm not breaking anything.

If I had a concern to raise is that I'm not 100% certain that nothing but human clients hits that API. If we have ATC or workers reaching out to this, it would mean provisioning certs for them too. There was also a different approach discussed in concourse/concourse#578 , but I personally think it's better practice to do blanket mTLS over limiting it to certain paths.

I'm not sure I understand, why this is written as an RFC and not simply a PR? This proposal does not change anything fundamental to Concourse's approach to authentication. I'm not sure this requires an RFC.

nickhyoti:
Hi,

I ended up doing this as an RFC as we weren't sure about the process.  

With respect to the scope, we only wanted to lock down the interface between a proxy and concourse to avoid issues around proxies being misconfigured and, for example, test and production systems end up getting interconnected in error.

vito:
@nickhyoti Thanks for doing a PR, and sorry for the confusion! I've updated the README to be much clearer around [when the RFC process is necessary](https://github.com/concourse/rfcs#when-the-rfc-process-is-necessary). I'll close this out since the feature has already been merged, and I don't think an RFC is needed for this. :)

PR #77, opened 2020-09-28
RFC: k8s storage

taylorsilva:
[Rendered](https://github.com/concourse/rfcs/blob/074-k8s-storage/074-k8s-storage/proposal.md)

This is meant for the next iteration of our Kubernetes runtime work. **This has nothing to do with how the Concourse helm chart currently functions**.

### DIFF
--- a/074-k8s-storage/proposal.md
+++ b/074-k8s-storage/proposal.md
@@ -72,9 +72,13 @@ When implementing a CSI driver for Kubernetes it is helpful to understand when c
 
 ## Proposed Implementation of Baggageclaim as a CSI Driver
 
+Targeting Kubernetes Version 1.19
+
 Follow the recommended deployment strategy from the Kubernetes team [described in this design document](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/container-storage-interface.md#recommended-mechanism-for-deploying-csi-drivers-on-kubernetes) with the following differences:
-- no `external-resizer` container
-- no `external-snapshotter` container
+- no `external-resizer` container. Not planning to support resizing.
+- no `external-snapshotter` container. We will use the `CLONE_VOLUME` feature to create COW volumes in baggageclaim instead of trying to use snapshots.
+- An extra volume must be mounted for each Pod in the DaemonSet. This volume, which should be very large, will be used by baggageclaim to store the volumes that it creates on each Kubernetes node.
+- We plan to not guarantee the requested storage capicity because we have no idea how much space any given step in Concourse will use. Kubernetes will force us to specify a storage request but our CSI driver will ignore this value.
 
 
 # Storage Options considered
--- a/074-k8s-storage/proposal.md
+++ b/074-k8s-storage/proposal.md
@@ -78,15 +78,15 @@ Targeting Kubernetes Version 1.19
 
 Follow the recommended deployment strategy from the Kubernetes team [described in this design document](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/storage/container-storage-interface.md#recommended-mechanism-for-deploying-csi-drivers-on-kubernetes) with the following differences:
 - no `external-resizer` container. Not planning to support resizing.
-- no `external-snapshotter` container. We will use the `CLONE_VOLUME` feature to create COW volumes in baggageclaim instead of trying to use snapshots.
+- no `external-snapshotter` container. The baggageclaim CSI driver will use the `CLONE_VOLUME` feature to create COW volumes in baggageclaim instead of trying to use snapshots.
 - An extra volume must be mounted for each replica Pod in the DaemonSet. This volume, which should be very large, will be used by baggageclaim to store the volumes that it creates on each Kubernetes node.
-- We plan to **not guarantee** the requested storage capicity because we have no idea how much space any given step in Concourse will use. Kubernetes will force us to specify a storage request but our CSI driver will ignore this value. This goes against the CSI spec.
+- The baggageclaim CSI drver will **not guarantee** the requested storage capicity because Concourse has no idea how much space any given step in will use, therefore Concourse continues to expect the CSI driver to follow this assumption. Kubernetes will force us to specify a storage request but the baggageclaim CSI driver will ignore this value. This goes against the CSI spec.
 
 Let's go over some use cases to get an understanding about how the implementation may work.
 
 ### Creating An Empty Volume
 
-A user creates a PVC:
+Concourse creates a PVC:
 ```yaml
 apiVersion: v1
 kind: PersistentVolumeClaim
@@ -104,7 +104,7 @@ spec:
 
 The [`external-provisioner`](https://github.com/kubernetes-csi/external-provisioner) will call `Controller.CreateVolume`. In this case `CreateVolume` will generate an ID for tracking the volume.
 
-With the PVC "created" (from the perspective of Kubernetes), a user can now reference the PVC in a Pod.
+With the PVC "created" (from the perspective of Kubernetes), Concourse can now reference the PVC in a Pod.
 
 ```yaml
 apiVersion: v1
@@ -155,6 +155,7 @@ The volume has been successfully provided to Kubernetes by this point.
 
 ### Creating A Cloned Volume
 
+Concourse creates a PVC that references another PVC in the `spec.dataSource` field:
 ```yaml
 apiVersion: v1
 kind: PersistentVolumeClaim
@@ -168,7 +169,7 @@ spec:
     requests:
       storage: 1Gi
   storageClassName: baggageclaim
-  dataSource: # will only support cloning other baggageclaim volumes
+  dataSource:
     name: some-other-pvc
     kind: PersistentVolumeClaim
 ```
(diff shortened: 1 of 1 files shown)


074-k8s-storage/proposal.md
```diff
@@ -0,0 +1,167 @@
+# Terms
+- **cache object** a BLOB and relevant metadata that Concourse needs to persist. These could be Resource Caches, Task Caches or Build Caches.
+- **worker** Concourse executes steps on a **worker** and implements some **worker** interface. Concourse is agnostic of the runtime representation of the worker (eg. K8s pod, node or cluster).
+
+# Summary
+
+After spiking on a few solutions for storage on Kubernetes our recommendation is to use an image registry to store **cache objects** for steps.
```
cirocosta, line 7: it might be interesting to check out the approach that [kaniko](https://github.com/GoogleContainerTools/kaniko) took - it essentially caches layers in a container registry if you don't specify a local path (so, e.g., if you want to run a `Job` that runs `kaniko` but you don't want to care about providing it a volume - which would imply `concurrency=1`)

```console
$ k logs -f build-v1.18.6-4tn7w  | grep cache
INFO[0008] Checking for cached layer cirocosta/envtest-cache:8642ec4096d286832017c667e68548ac74436793bd49b1cf59d6a238044f6b0a...
INFO[0008] No cached layer found for cmd RUN set -x &&          apt update -y && apt install -y rsync &&                git clone https://github.com/kubernetes/kubernetes &&                 cd ./kubernetes &&                      git checkout $KUBE_TAG &&            go mod download &&                       make generated_files
INFO[0194] Checking for cached layer cirocosta/envtest-cache:21931b04c009b9e67db0c6c8a1f907ee699dd32ad3fef2aa14efaf6b1a5dc995...
INFO[0313] Found cached layer, extracting to filesystem
INFO[0348] Checking for cached layer cirocosta/envtest-cache:b862d7fa29205a55ec028679dc2bb416b707a1a29d24feb9714075e4be5a5c74...
INFO[0435] Found cached layer, extracting to filesystem
INFO[0447] Checking for cached layer cirocosta/envtest-cache:e6c3ee315be35a8483933a79682375de049bdef0560721ce9962d728993803d0...
INFO[0464] Found cached layer, extracting to filesystem
INFO[0470] Checking for cached layer cirocosta/envtest-cache:e78396635e5a8c3a4b750b11796109b2a6a954b628c6e24a3dfbc1f579ce7024...
INFO[0471] No cached layer found for cmd RUN go install -v ./cmd/envtest
INFO[0508] Pushing layer cirocosta/envtest-cache:e78396635e5a8c3a4b750b11796109b2a6a954b628c6e24a3dfbc1f579ce7024 to cache now
```

(https://github.com/GoogleContainerTools/kaniko/blob/master/pkg/cache/cache.go)

074-k8s-storage/proposal.md
```diff
@@ -0,0 +1,167 @@
...
+### Pros
++ Leverage baggageclaim
+	+ volume streaming between nodes would work using the current Concourse architecture
+	+ resource and task caches would also work using the current Concourse architecture
+	+ would be able to stream **cache objects** across worker runtimes as it would be mediated via the web
++ Concourse would have complete control over volume lifecycle
++ would have negligible overhead for steps scheduled on the same node as no input/output stream would be required
+
+### Cons
+- Not being able to use IaaS based persisent disks doesn't offer a viable solution. K8s nodes would need to have large root volumes.
```
ari-becker, line 50: Is this actually an issue? Under `Persistent Volumes` you note as a con:
> IaaS based limits on [volume limits per node](https://kubernetes.io/docs/concepts/storage/storage-limits/#dynamic-volume-limits) prevents this from being a scalable solution
Large root volumes is closer to how Concourse Workers work today (i.e. closer to the implicit foundational design assumptions in the Concourse domain that this RFC attempts to deal with), and if there are scalability issues with Persistent Volumes, then doesn't this represent a simpler way (because it builds on prior `baggageclaim` work) to get started with building the Kubernetes runtime?

074-k8s-storage/proposal.md
```diff
@@ -0,0 +1,167 @@
...
+- stream **cache objects** across worker runtimes (k8s worker sends artifact to garden worker)
+
+## Criteria
+- security
+- performance, aka initialization time (time spent running workloads on a single k8s worker, as well as across workers)
+- resource usage to run this storage solution
+
+# Proposal
+
+**TL;DR**: We recommend going with the image registry option because it satisfies all the requirements and gives us a bunch of options to improve performance when compared to the blobstore option. It also provides a very flexible solution that works across multiple runtime workers. [See Details](#image-registry-to-store-artifacts)
```
xtreme-sameer-vohra, line 28: Trade-study - push/pull from registry using
1. init & side car containers
1. CSI with independent volume/input (eg. image populator)
1. Custom image per combination of step for pulling and TODO for pushing

074-k8s-storage/proposal.md
xtreme-sameer-vohra: Hey @cirocosta 
We can see this style of local caching being beneficial for pulling the contents on an input for a task. Is that what you meant ? We weren't sure how this might be useful for pushing 🤔

074-k8s-storage/proposal.md
```diff
@@ -0,0 +1,167 @@
...
+	+ resource and task caches would also work using the current Concourse architecture
+	+ would be able to stream **cache objects** across worker runtimes as it would be mediated via the web
++ Concourse would have complete control over volume lifecycle
++ would have negligible overhead for steps scheduled on the same node as no input/output stream would be required
+
+### Cons
+- Not being able to use IaaS based persisent disks doesn't offer a viable solution. K8s nodes would need to have large root volumes.
+- Wouldn't have support for hosting images by default. However, `baggageclaim` could be extended to add the APIs
+- `baggageclaim` itself doesn't have any authentication/authorization or transport security (https) mechanisms built into it
+
```
xtreme-sameer-vohra, line 53:
```suggestion
+ - `baggageclaim` needs to be run in a privileged container. It would be useful to understand the perspective of operators for this constraint ? Is this tenable ?
```

074-k8s-storage/proposal.md
xtreme-sameer-vohra: Hey @ari-becker,
Yes, this is true. It would be useful to get perspectives such as yours for us to understand what constraints or lack thereof folks are dealing with. 

We are assuming the cons;
- resizing of nodes with larger root volumes
- running privileged baggageclaim pods
- being ok with baggagelciam's current lack of security (this can be enhanced)
creates larger challenges for operators.

It sounds like, you'd be okay with this setup ?

ari-becker: @xtreme-sameer-vohra currently, we already run Concourse workers on a Kubernetes cluster, one where we've architected our workers such that the workers are restricted to their own underlying nodes (we manually scale both the underlying nodes and the worker StatefulSet in tandem) and the worker pods are privileged. So all of your cons are *already present* in our deployment and therefore implicitly acceptable.

Maybe you need to clarify what the goals are supposed to be of developing a Kubernetes runtime for Concourse? is it to phase out the "worker" in favor of ATC directly scheduling Kubernetes Jobs that represent Concourse jobs or steps? Because then running privileged baggageclaim pods are fine, basically then the worker isn't really phased out, it's just simplified down to only the baggageclaim code, and presumably we'd be able to configure ATC to create Jobs with node restrictions / affinity / taints / tolerations to get the Jobs to run on a dedicated worker instance group. Is it the ability to run on common / shared instances? Because then running privileged baggageclaim pods is probably unacceptable. Should untrusted Concourse workloads run on common / shared instances in the first place, even if they're not allowed to be privileged? Ehhh, debatable, and a lot of security experts will tell you no, particularly if the jobs are not configured with PodSecurityPolicies and the like, which in a CI / development environment rarely makes sense.

The main value that we'd get out of a Kubernetes-native runtime would be the ability to set up [Cluster Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) so that Kubernetes will add additional nodes (and reduce nodes) as needed according to scheduling workloads, as currently we need to scale workers by hand. Cluster Autoscaler works well with workloads that are scheduled to dedicated instance groups, adding and subtracting nodes from the dedicated instance group that will permit Jobs that are restricted to those instance groups.

xtreme-sameer-vohra: Ah yes, we are working on the [K8s Runtime RFC](https://github.com/concourse/rfcs/blob/075-k8s-runtime/075-k8s-runtime/proposal.md) that should have preceded this one 😸 
The primary objective being able to leverage the scaling & operational capabilities of K8s by deploying `steps` as K8s pods without changing the pipeline schema. The latter would allow pipelines to be portable across runtimes. Along the same lines we're also assessing;
- the least amount of required permissions on K8s for fulfilling the same requirements as a current worker (container & storage management)
- the # of dependencies (besides postgres) for a Concourse deployment on K8s

As you mentioned, the container management would be delegated to k8s. We are [spiking](https://github.com/concourse/concourse/issues/6133) on using IaaS PVs + propagating mounts using CSI, hoping to ultimately leverage `baggageclaim`. We are early in the exploration, but this path might also provide the ability of using volume location for the purposes of scheduling 💭 

The initial draft of the proposal has a worker being mapped to K8s cluster + namespace. "Instance Groups" seems interesting. Do you see it fitting in naturally with the cluster+namespace mapping or would you propose something different ?

ari-becker: @xtreme-sameer-vohra One of the things that worries me about using an external image registry is the prospect of cross-availability-zone (on AWS, $0.01/GB) and/or VPC egress traffic costs (on AWS, much higher). Currently, all of our workers are in the same availability zone (for CI workers, we're less concerned with the prospect of an AZ-wide outage and more concerned with the traffic costs) and so our traffic costs are much lower.

Effectively using a external image registry while keeping costs under control would mean a) deploying a separate image registry per availability zone alongside the workers, where workers are correctly configured to use the image registry that is in their availability zone, b) the image registry itself needs to use either a local disk / IaaS PV (so as not to incur traffic costs when storing layers in s3, for example), so really all that an image registry does is consolidate storage usage in one place. This eventually leads to cost savings compared to using a large local disk / IaaS PV per worker, with the tradeoff of additional operational complexity.

When talking about instance groups, I'm using the [Kops](https://github.com/kubernetes/kops/blob/master/docs/instance_groups.md) nomenclature. An instance group maps to an AWS AutoscalingGroup, where the nodes are of a certain instance type (i.e. CI is generally best-served with burstable nodes, which on AWS are t3 nodes, whereas t3 nodes may not make sense for other workloads). In order to schedule workers on a specific ASG that we set up for them, we use a combination of [node affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity) to direct Kubernetes to schedule the CI workloads (currently, the worker pods) to those nodes, and [taints and tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) to not allow any other workloads to schedule there. This is orthogonal to Kubernetes namespacing, which is a good idea anyway to improve logical isolation between individual CI workloads.

xtreme-sameer-vohra:
TODO: Add a section explaining the current storage mechanism and where it falls short

074-k8s-storage/proposal.md
cirocosta: as far as i know, TKG clusters created via TMC expects pods to adhere to a podsecurity policy that's quite restrictive:

- https://docs.vmware.com/en/VMware-vSphere/7.0/vmware-vsphere-with-tanzu/GUID-CD033D1D-BAD2-41C4-A46F-647A560BAEAB.html
- https://docs.pivotal.io/tkgi/1-9/pod-security-policy.html

buuut, I'd imagine someone installing a CSI plugin (like anything network related) would expect to grant higher privileges anyway (being an "infra component")

074-k8s-storage/proposal.md
```diff
@@ -0,0 +1,172 @@
...
+Each **cache object** would be stored in its own persistent volume. Persistent volume snapshots would be used to reference **cache object** versions.
+
+### Pros
+- Would leverage native k8s offering
+- Maps well to Concourse's use of **cache objects** and offloads the heavy lifting to K8s
+- Potentially wouldn't require volumes to be streamed at all
+
+### Cons
+- Wouldn't have support for hosting images by default.
+- IaaS based limits on [volume limits per node](https://kubernetes.io/docs/concepts/storage/storage-limits/#dynamic-volume-limits) prevents this from being a scalable solution
```
cirocosta, line 114: not super related, but found this idea interesting: https://github.com/argoproj/argo/issues/4130

074-k8s-storage/proposal.md
cirocosta: oh, I didn't have any specifics in mind, it was more about showcasing the pattern of using a registry as a place where you just throw layers at a registry regardless of what you're building and then try to reuse those layers (pulling if they exist) when running another build

074-k8s-storage/proposal.md
xtreme-sameer-vohra: Yep, this does make the assumption that installing Concourse requires more than just pod* capabilities. 
However, installing a `CSI driver` would be an `Operator` action similar to installing concourse today.  

@taylorsilva and I were also discussing that this approach wouldn't preclude us from using the IaaS PVs directly for simpler installations where user doesn't have privileges to install a `CSI driver`. This would have performance implications in the overhead to create/destroy volumes and limit of volumes / node, but would be a very simple solution 🤔

xtreme-sameer-vohra:
TODO - Document the permissions required to setup the baggageclaim CSI + Storage Class

taylorsilva:
We are not doing this anymore. All runtime efforts will focus on containerd now.

PR #82, opened 2020-10-21
P2P Volume Streaming

evanchaoli:

### DIFF
--- a/082-p2p-volume-streaming/proposal.md
+++ b/082-p2p-volume-streaming/proposal.md
@@ -1,0 +1,99 @@
+# Summary
+
+Introduces P2P Volume Streaming, an enhancement of volume streaming that will
+benefit clusters where workers are reachable from each other.
+
+# Motivation
+
+When Concourse launches a container on a worker, if some required volumes locate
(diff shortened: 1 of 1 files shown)


evanchaoli:
I have done a PoC with performance comparison with and without P2P Volume Streaming, see https://github.com/concourse/concourse/pull/6186#issue-507204153.

xtremerui:
Hi, @evanchaoli I think we could consider this one merged (though some implementations are already done) if we could sort out how P2P streaming works for worker within a different network i.e. external worker that under a completely different network. 

I have read your comments under related PRs. Could you have a brief summary of those comments and update this RFC with it so we could proceed? Thx!

evanchaoli:
@xtremerui I have updated the RFC, please take a look.

xtremerui:
one thing is unclear to me, if all workers are not in the same network and P2P streaming is enabled, what is the expected behaviour? I think we should make it clear in the feature flag description about the consequence of doing that.

evanchaoli:
@xtremerui Just add a statement in "New implications", please take a look.

PR #97, opened 2021-03-17
Design principles + README update

vito:
This PR introduces a `DESIGN_PRINCIPLES.md` document which codifies the set of principles underlying Concourse's feature design sensibilities. These principles have been referenced by the [v10 blog post](https://blog.concourse-ci.org/core-roadmap-towards-v10/) and many of my comments on GitHub so hopefully they don't feel too surprising.

Along the way, this PR updates `README.md` to reference the design principles and the new process by which RFCs are placed directly on the [Concourse website](https://concourse-ci.org). (Which hasn't been shipped yet - I'll make this PR a non-draft once it's ready to be merged.)

### DIFF
--- a/DESIGN_PRINCIPLES.md
+++ b/DESIGN_PRINCIPLES.md
@@ -21,4 +21,4 @@ Concepts should not be highly specialized for one domain or introduce tight coup
 
 Concourse should prevent [anti-patterns](https://github.com/concourse/concourse/wiki/Anti-Patterns) and the accumulation of technical debt. Concourse's concepts should make good practices feel intuitive and bad practices feel uncomfortable.
 
-Automation should be portable and reproducible. Recovery from disaster scenarios (e.g. total cluster loss) should be possible with minimal disruption to the path to production. Each piece of mission-critical state should be kept in an external source of truth, allowing Concourse to be an ephemeral choreographer.
+Automation should be self-contained and reproducible in order to maintain business continuity when recovering from disaster scenarios (e.g. total cluster loss). Concourse should merely be a choreographer of mission-critical state kept in an external source of truth, allowing individual installations to be ephemeral.
--- a/DESIGN_PRINCIPLES.md
+++ b/DESIGN_PRINCIPLES.md
@@ -21,4 +21,4 @@ Concepts should not be highly specialized for one domain or introduce tight coup
 
 Concourse should prevent [anti-patterns](https://github.com/concourse/concourse/wiki/Anti-Patterns) and the accumulation of technical debt. Concourse's concepts should make good practices feel intuitive and bad practices feel uncomfortable.
 
-Automation should portable and reproducible, able to be recovered with minimal disruption to the path to production. All mission-critical state should be kept external to Concourse, leaving Concourse to just be a connector-of-dots.
+Automation should portable and reproducible, able to be recovered with minimal disruption to the path to production. Each piece of mission-critical state should be kept in an external source of truth, allowing Concourse to be an ephemeral choreographer.
--- a/DESIGN_PRINCIPLES.md
+++ b/DESIGN_PRINCIPLES.md
@@ -5,7 +5,7 @@ Concourse's goal is to solve automation, once and for all, without becoming part
 
 ## Expressive by being precise
 
-Concourse should provide concepts that build a strong mental model for the user's project that remains intuitive as their automation grows.
+Concourse should provide concepts that build a strong mental model for the user's project that remains intuitive as their automation requirements grow.
 
 Concepts should precisely outline their motivation and intended workflows. Friction and complexity resulting from the imprecise application of a concept should be a cue to introduce new ideas. ([Example](https://blog.concourse-ci.org/reinventing-resource-types/))
 
@@ -21,4 +21,4 @@ Concepts should not be highly specialized for one domain or introduce tight coup
 
 Concourse should prevent [anti-patterns](https://github.com/concourse/concourse/wiki/Anti-Patterns) and the accumulation of technical debt. Concourse's concepts should make good practices feel intuitive and bad practices feel uncomfortable.
 
-Automation should portable and reproducible, able to be recovered with minimal disruption to the path to production. Each piece of mission-critical state should be kept in an external source of truth, allowing Concourse to be an ephemeral choreographer.
+Automation should be portable and reproducible. Recovery from disaster scenarios (e.g. total cluster loss) should be possible with minimal disruption to the path to production. Each piece of mission-critical state should be kept in an external source of truth, allowing Concourse to be an ephemeral choreographer.
--- a/DESIGN_PRINCIPLES.md
+++ b/DESIGN_PRINCIPLES.md
@@ -1,0 +1,24 @@
+# Concourse Design Principles
+
+Concourse's goal is to solve automation, once and for all, without becoming part of the problem.
+
+
+## Expressive by being precise
+
+Concourse should provide concepts that build a strong mental model for the user's project that remains intuitive as their automation grows.
+
+Concepts should precisely outline their motivation and intended workflows. Friction and complexity resulting from the imprecise application of a concept should be a cue to introduce new ideas. ([Example](https://blog.concourse-ci.org/reinventing-resource-types/))
+
+
+## Versatile by being universal
+
+Concourse should be able to do a lot with a little. New concepts should only be introduced if their intended workflow cannot be precisely expressed in terms of existing concepts.
+
+Concepts should not be highly specialized for one domain or introduce tight coupling to specific technologies. Users should be able to relate to every concept, and their automation should be insulated from the constant churn of the tech industry.
+
+
+## Safe by being destructible
+
+Concourse should prevent [anti-patterns](https://github.com/concourse/concourse/wiki/Anti-Patterns) and the accumulation of technical debt. Concourse's concepts should make good practices feel intuitive and bad practices feel uncomfortable.
+
+Automation should portable and reproducible, able to be recovered with minimal disruption to the path to production. All mission-critical state should be kept external to Concourse, leaving Concourse to just be a connector-of-dots.
--- a/DESIGN_PRINCIPLES.md
+++ b/DESIGN_PRINCIPLES.md
@@ -5,7 +5,7 @@ Concourse's goal is to solve automation, once and for all, without becoming part
 
 ## Expressive by being precise
 
-Concourse should provide concepts that build a strong mental model for the user's project that remains intuitive as their automation requirements grow.
+Concourse should provide concepts that build a strong mental model for the user's project, and this model should remain intuitive as the user's automation requirements grow.
 
 Concepts should precisely outline their motivation and intended workflows. Friction and complexity resulting from the imprecise application of a concept should be a cue to introduce new ideas. ([Example](https://blog.concourse-ci.org/reinventing-resource-types/))
 
--- a/README.md
+++ b/README.md
@@ -155,6 +155,20 @@ From there, the implementation process falls under the normal [Concourse
 development process][contributing].
 
 
+## Revising an RFC
+
+RFCs represent the planning phase. An RFC's proposal is not the source of truth
+for the feature's documentation, and should not be revised to keep up with
+later iterations after the initial proposal is implemented. A new RFC should be
+proposed for subsequent changes instead.
+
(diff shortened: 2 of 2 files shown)


DESIGN_PRINCIPLES.md
```diff
@@ -0,0 +1,24 @@
...
+Concourse should be able to do a lot with a little. New concepts should only be introduced if their intended workflow cannot be precisely expressed in terms of existing concepts.
+
+Concepts should not be highly specialized for one domain or introduce tight coupling to specific technologies. Users should be able to relate to every concept, and their automation should be insulated from the constant churn of the tech industry.
+
+
+## Safe by being destructible
+
+Concourse should prevent [anti-patterns](https://github.com/concourse/concourse/wiki/Anti-Patterns) and the accumulation of technical debt. Concourse's concepts should make good practices feel intuitive and bad practices feel uncomfortable.
+
+Automation should portable and reproducible, able to be recovered with minimal disruption to the path to production. All mission-critical state should be kept external to Concourse, leaving Concourse to just be a connector-of-dots.
```
matthewpereira, line 24: "Automation should **be** portable and reproducible, **and recovery should be possible** with minimal disruption to the path to production. All mission-critical state should be kept external to Concourse, leaving Concourse to just be a connector-of-dots."

Should we consider 'choreographer' or 'orchestration' instead of 'connector-of-dots'? 🤔

matthewpereira (changes requested):

DESIGN_PRINCIPLES.md
vito: lol yes probably

vito: Though there's also some aspect of statelessness I'd like to convey there - the dots metaphor comes from [this review comment](https://github.com/concourse/rfcs/pull/36#pullrequestreview-336557490) which made a point that Concourse itself is not a 'dot'.

vito: pushed another edition! lemme know what you think

aoldershaw: looks like you're still missing the corrections to the first sentence outline in https://github.com/concourse/rfcs/pull/97#discussion_r596038405 (esp. adding `be` to `Automation should ... portable`)

vito: woops, didn't even notice the bold text there - fixing!

README.md
```diff
@@ -3,82 +3,95 @@
...
-of comments at the top level.
+An RFC may not be necessary for changes that are narrow enough in scope and
+impact. If you feel that this is the case, you can jump straight to submitting
+a PR. Do note however that pull requests may be closed with a polite request to
+submit an RFC if we find that it needs broader feedback.
 
-Top-level comments and pull-request reviews are allowed for overarching
-commentary, but in general line-wise comments are preferred.
+If you're not sure whether to open an RFC for a change you'd like to propose,
+feel free to [ask in `#dev`](https://discord.gg/MeRxXKW).
```
aoldershaw, line 37: Unrelated to this PR, but access to `#dev` requires being a contributor - is there anything on Discord that indicates how to become a contributor? Should we add something to the channel description or is there a better way?

aoldershaw (approved):

README.md
vito: The `#welcome` channel does but it's out of date (just says ask for it) - we can add a link to https://github.com/concourse/governance in the topic.

DESIGN_PRINCIPLES.md
```diff
@@ -0,0 +1,24 @@
+# Concourse Design Principles
+
+Concourse's goal is to solve automation, once and for all, without becoming part of the problem.
+
+
+## Expressive by being precise
+
+Concourse should provide concepts that build a strong mental model for the user's project that remains intuitive as their automation requirements grow.
```
matthewpereira, line 8: Suggestion:
"Concourse should provide concepts that build a strong mental model for the user's project, **and this model should remain** intuitive as **the user's** automation requirements grow.

README.md
```diff
@@ -3,82 +3,95 @@
...
-architecture and feature design process, before getting to code and
-implementation.
+Everyone is encouraged to provide feedback on RFCs!
 
-An RFC may not be necessary for changes that are narrow enough in scope with
-limited impact to the rest of Concourse. If you feel that this is the case, you
-can cut straight to submitting a PR, though it's still a good idea to have an
-issue opened first to provide additional context. Do note however that pull
-requests and issues may be closed with a polite request to submit an RFC first.
+RFC review should be focused on bringing clarity to the proposal, resolving
```
matthewpereira, line 10: RFC review should **focus on** bringing clarity to the proposal, resolving

matthewpereira (approved):
@vito I added a pair of suggestions to make a few sentences a little clearer/easy to read, but they're optional/non-blocking. Otherwise LGTM! :shipit:

DESIGN_PRINCIPLES.md
```diff
@@ -0,0 +1,24 @@
...
+Concourse should be able to do a lot with a little. New concepts should only be introduced if their intended workflow cannot be precisely expressed in terms of existing concepts.
+
+Concepts should not be highly specialized for one domain or introduce tight coupling to specific technologies. Users should be able to relate to every concept, and their automation should be insulated from the constant churn of the tech industry.
+
+
+## Safe by being destructible
+
+Concourse should prevent [anti-patterns](https://github.com/concourse/concourse/wiki/Anti-Patterns) and the accumulation of technical debt. Concourse's concepts should make good practices feel intuitive and bad practices feel uncomfortable.
+
+Automation should be portable and reproducible. Recovery from disaster scenarios (e.g. total cluster loss) should be possible with minimal disruption to the path to production. Each piece of mission-critical state should be kept in an external source of truth, allowing Concourse to be an ephemeral choreographer.
```
chenbh, line 24: Is portability still a core tenet? I remember there were conversations about how even though Concourse tries to prevent vender lock in via separation of pipelines and scripts, it's still very difficult to rip Concourse out of an existing infrastructure.

In addition, I'm not sure how much this principle has actually affected RFCs, I can technically argue that any enhancements to pipelines violates portability since it'll make it harder for me switch off Concourse. It might better be expressed as a clean separation between infrastructure config (pipeline/task), and code (script).

vito: Portability in this context doesn't mean portability between Concourse and other CI systems - it just means the ability to easily migrate between Concourse clusters if needed, possibly running different versions (to the extent that needed features are supported) or being maintained by different operators with different configurations.

For example, this principle is what leads us to prefer that pipelines be as self-contained as possible, configuring their own dependent resource types rather than having them be configured cluster-wide.

chenbh: Ah, can we clarify that, maybe something like:

```suggestion
Automation should be portable and reproducible. Recovery from disaster scenarios (e.g. total cluster loss) should be possible with minimal disruption to the path to production. Each piece of mission-critical state should be kept in an external source of truth, allowing individual Concourse clusters to be an ephemeral choreographer.
```

vito:
@concourse/core Think I'm finally done with the new README if y'all wanna give it a gander!

DESIGN_PRINCIPLES.md
vito: How's this?

```suggestion
Automation should be self-contained and reproducible in order to maintain business continuity when recovering from disaster scenarios (e.g. total cluster loss). Concourse should merely be a choreographer of mission-critical state kept in an external source of truth, allowing individual installations to be ephemeral.
```

README.md
```diff
@@ -1,132 +1,181 @@
...
 
-A process for collaborating on substantial changes to Concourse.
+The Concourse project uses the RFC (request for comments) process for
+collaborating on substantial changes to Concourse. RFCs enable contributors to
+collaborate during the design process, providing clarity and validation before
+jumping to implementation.
 
 
-## What should be proposed as an RFC?
+## When this process is necessary
```
matthewpereira, line 9:
```suggestion
## When the RFC process is necessary
```

README.md
```diff
@@ -1,132 +1,181 @@
...
+RFCs represent the planning phase. An RFC's proposal is not the source of truth
+for the feature's documentation, and should not be revised to keep up with
+later iterations after the initial proposal is implemented. A new RFC should be
+proposed for subsequent changes instead.
+
+If an RFC is merged and later changes are deemed necessary prior to final (i.e.
+non-experimental) implementation, a follow-up PR may be submitted that updates
+the proposal in-place. In this case the RFC author must include a MAJOR.MINOR
+revision number in the proposal and maintain a brief summary of changes at the
+bottom of the proposal.
```
chenbh, line 169: Is there a procedure for deprecating / obsoleting an RFC? Let's say `spatial-resources` was merged in as an RFC, but then we came out with `resource-prototypes`, how to we mark it as "superseded" (or "partially superseded")? And who's responsibility is it to make the change to the old RFC (I'm guessing it's the new RFC's author)?

chenbh: I think I should've created this comment on a newline instead of on that particular paragraph, but I was more concerned with what's the procedure for after RFC is merged and the final implementation is done. Let's say `040-valid-identifiers` is merged (which it currently is), and has gone GA'd (which it currently isn't).

1. Is there a procedure for updating the proposal (e.g. `~` is now a valid character)? Does it need to go through the experimental implementation phase once again? Or maybe do we just submit a *new* RFC and mark 040-valid-identifiers as `replaced by: ###-valid-identifiers-v2`?
2. What happens if there's a new RFC that partially replaces it? Let's say we have a new RFC `###-valid-job-syntax` which overlaps (e.g. stricter identifiers for jobs) but also contains additional content (e.g. reserved words, structure, etc). What's the wording for "Some part of this RFC has been supplemented by RFC ###"?

vito: Those are good questions, but I think they may indicate that we have two different perspectives of RFCs getting tangled up between our comments:

* RFCs serve as a collaboration tool for planning changes to a single project.
* RFCs serve as authoritative design documents for the broader ecosystem to rally around and implement (e.g. open standards and protocols like SSH).

Both are perfectly valid but serve different use cases. What I'm trying to say is that our RFC process is currently designed for the former, not the latter. We have quite a few proposals that seem like the latter, but that's because they introduce entirely new concepts (e.g. prototypes, projects) or define internal standards for the first time (e.g. valid identifiers). I expect later iterations to be done through smaller followup RFCs, and for the Concourse documentation to remain the authoritative description of the current state of the world.

For example, imagine someone joins the project 5 years from now and wants to make `*` valid in identifiers. I wouldn't expect them to go through all previous RFCs to find anything that mentions identifiers so they can reference it. I also wouldn't expect them to introduce a new 'valid identifiers' proposal to replace the old one, because that would make it harder for reviewers to know what changes are actually being proposed (i.e. what's the delta).

So, if the `040-valid-identifiers` RFC was completed and GA'd, it would then just become part of the baseline for any future RFCs, and the `040-valid-identifiers` document should be effectively frozen. Any followup changes to identifiers would need to be submitted as a new proposal, and the new proposal wouldn't even need to be aware of the previous one.

The RFC author may choose to link to prior work in the motivation/summary/context/etc., but I don't think the reverse is necessary, and I don't think we need to encode a process for either. Over time managing these links may lead to high overhead, and it's likely for connections to be missed eventually as older proposals become common assumptions, making the whole process kind of moot.

Reviewers may bring up overlap with other proposals and request clarification, but I don't know if we need a special procedure for that either - there are probably lots of things reviewers might bring up, and I'd prefer to err on the side of being too lightweight rather than too prescriptive, at least until we see how things play out.

Let me know if this makes sense!

README.md
```diff
@@ -1,132 +1,181 @@
...
+if you don't have much more to add.
 
-* `resolution/merge`: the PR will be merged!
-* `resolution/postpone`: the PR will be closed and stamped with a `postponed`
-  label. At some point in the future the pull request may be re-opened.
-* `resolution/close`: the PR will be closed with no additional label.
+
+### Resolution
+
+When enough clarity and consensus has been reached the core team assignee will
```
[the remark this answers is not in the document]
vito: clarified in https://github.com/concourse/rfcs/pull/97/commits/f0093b2ca06b69b7962d8b4843e394ed0436faa1!

(ran it by @clarafu in-person; resolving thread)

README.md
chenbh: 👍  I was indeed thinking of RFCs more like the IETF living-standards style

vito:
Thanks for the feedback all! Went over it one more time and I think it's ready to merge; gonna hit the big green button now.

PR #110, opened 2021-07-27
RFC: shared webhooks

aoldershaw:
[Rendered](https://github.com/aoldershaw/rfcs/blob/shared-webhooks/110-shared-webhooks/proposal.md)

### DIFF
--- a/110-shared-webhooks/proposal.md
+++ b/110-shared-webhooks/proposal.md
@@ -1,0 +1,169 @@
+# Summary
+
+Decouple webhooks from individual resources, allowing Concourse to process a webhook payload and queue a `check` for the approriate resource(s) (and eventually [var sources]).
+
+
+# Motivation
+
+Concourse's primary method of detecting new resource versions is through periodically running `check`. The default polling frequency for each resource is 1 minute.
+
+Polling can lead to a lot of pressure on the other end of the `check` requests. This can lead to rate limiting, or worse, your CI/CD stack DDoSing your own internal services.
+
+To reduce polling pressure it is typical to configure webhooks, by configuring a `webhook_token` on a resource and either setting a higher `check_every` or configuring `--resource-with-webhook-checking-interval` cluster-wide. This token value is then given to the service, as part of a per-resource URL.
+
+This is a tedious operation and has severe limitations, as some external services (e.g. GitHub) have a limit on how many webhooks can be configured on a repo/org. This means you can only configure webhooks for up to 20 resources (minus whatever other hooks you have), which is a painful limitation when there's a common resource used in many pipelines.
+
+Additionally, especially when operating a large Concourse deployment spanning many different teams, it's difficult to get users to configure webhooks since it requires per-resource configuration (both in and out of Concourse). It would be much simpler to define webhooks globally and have them automatically apply to matching resources.
+
+It is also infeasible to use webhooks with dynamic groups of [instanced pipelines]. For instance, if you set an instanced pipeline for each GitHub pull request, you would need to somehow update your webhook configuration within GitHub to point to the correct resources.
+
+Relevant issues/discussions:
+* https://github.com/concourse/concourse/issues/2230
+* https://github.com/concourse/concourse/discussions/6869#discussioncomment-949051
+
+
+# Proposal
+
+I propose adding a new webhook entity that can be managed via `fly`:
+
+```sh
+$ fly set-webhook --name github --type github --token abc123def456 (--team my-team | --global)
+url: https://ci.concourse-ci.org/api/v1/teams/my-team/webhooks/github?token=abc123def456
+```
+
+...and a `webhooks` field to be added to the [prototype info response] that tells Concourse what webhook payload(s) should trigger a check for the resource. For instance, the resource:
+
+```yaml
+resources:
+- name: my-repo
+  type: git # note: this is the git prototype, not resource_type
+  source:
+    uri: git@github.com:concourse/concourse.git
+    branch: master
+```
+
+may produce the following `info` response:
+
+```json
+{
+  "interface_version": "1.0",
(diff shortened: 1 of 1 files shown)


EstebanFS:
@taylorsilva

schmurfy:
Nice to see some movement on this front, I just read it and although I am not familiar enough with the way prototypes works it seems overly complicated. This may not be the type of feedback you wanted but since prototypes are in development since I started using concourse (~ 1.5 years ago) I imagine that following through with this proposal means that the current webhook system is going to be there for quite a while.

I really love concourse and this is by far the best automation system I have used so far but it each features takes years to develop it might become the ultimate automation tool at some point but in the meantime the temporary fixes become permanent and it just generate confusion for newcomers.

I am currently using `concourse-webhook-broadcaster` because there is not real way to use webhooks otherwise with dynamically generated pipelines (for pull-request, release, deployment).

110-shared-webhooks/proposal.md
```diff
@@ -0,0 +1,171 @@
...
+## Reducing `check` Frequency
+
+In order to reduce the polling pressure of resources that are backed by a webhook, we can keep track of when the last webhook check for a resource config completed.
+
+When we try to queue a check in Lidar, we check if has been checked by a webhook. If so, we could set the default check interval to something much larger (e.g. `24h`). The intent of configuring a high interval, rather than disabling the polling entirely, is to be resilient to missed events.
+
+
+# Open Questions
+
+### Can we avoid needing to implement `types` within Concourse?
```
ari-becker, line 132: One idea here would be to simply register webhooks on a team- or Concourse-install level, and then have a `webhook_ref` field under `resource` to register that resource with the global webhook handler.

Pro: doesn't require waiting for prototypes
Con: not as "clean" as automatic registration with prototypes

ari-becker (commented):
Excited to see this - we also hit the 20 webhook limit in GitHub, so this is really necessary.

110-shared-webhooks/proposal.md
aoldershaw: Yeah, that's what I was thinking originally. However, I feel the benefits of automatic registration are super valuable in large-scale deployments with many independent teams, where orchestrating a change to hundreds/thousands of pipelines may not be so easy.

That said, maybe we could support both automatic registration and manual registration via the pipeline config. That way, we can still see some benefits without needing to wait for prototype support

aoldershaw: BTW, even configuring the webhook manually via the pipeline config may still require implementing some `types` within Concourse (at least with how I'm envisioning it) - e.g. a slack webhook still needs special logic to respond to the webhook challenge.

ari-becker: Right, obviously there are many benefits to a prototype-driven model.

Honestly, my perspective right now is clouded by a current pipeline that I'm working on where I have a templated Terraform configuration that is currently instanced 4x with a current need to scale it to 100x and plans to scale it to 400-500x and beyond within a year, with the configuration per instance plus the template stored in the same repository. I want to have a job per instance, so that individual instances can be triggered when the configuration for an individual instance changes, without triggering the rest of the instance jobs unnecessarily. This would require different resources per instance, which doesn't scale with webhooks (as I will max out the webhook count on the repository when I hit about 20 instances), so currently I have a single resource for all of the instances which triggers every instance job. So it's rather inefficient.

aoldershaw: Sounds like a cool use-case, and exactly the kind that this RFC is motivated by! Some clarifying questions:
* Do all of these instances have configuration within the same repository?
* Are you looking to use a single webhook on the repository for all of these instances?
* Is each resource that's responding to these webhooks just tracking the config file for the single instance?

Assuming the answer to all of these questions is yes for the time being...

I was thinking it could look something like this:

```yaml
resources:
- name: config
  type: git
  source:
    uri: git@github.com:my-org/my-repo.git
    paths: [config/abc.yml]
  webhooks:
  - name: my-repo
    filter:
      commits:
      - modified: [config/abc.yml]
```

(I imagine a filter is important here, since otherwise every commit would trigger 400-500 checks)

ari-becker: Yes, yes, and no (but basically yes). Actually we were inspired by what https://github.com/concourse/governance did with keeping most of the data in YAML, so we have a layout like
```text
data/
  templates/
    instance.yaml
    role.yaml
  roles/
    chef.yaml
    programmer.yaml
    teacher.yaml
  instance/
    foo.yaml
    bar.yaml
tf/
  locals.tf
  example.tf
```
Where, for instance, `foo` (as defined inside the YAML) might be a `chef` and a `programmer` and `bar` might be a `programmer` and a `teacher`. Terraform compiles what the different roles are supposed to mean in terms of specific Terraform resources, and creating a new instance is as simple as `cp data/templates/instance.yaml data/instance/baz.yaml` and then re-generating the pipeline (probably I will eventually move it to an instanced pipeline, especially if this PR succeeds), where a specific instance of Terraform is isolated by running e.g. `terraform workspace select foo && terraform apply -var instance_id=foo`, which allows Terraform to arbitrarily scale while staying performant by not keeping too many instances in the same state.

aoldershaw: I took a crack at implementing the manual form of this RFC here: https://github.com/concourse/concourse/pull/7445

Only supports team-scoped webhooks, but would be easy to extend to global. Curious to hear your thoughts, and whether you feel it would suit your needs.

ari-becker: @aoldershaw Nice! I needed to go back and forth a little bit to see how it would filter the webhook events with `commits.modified` (it's not inherently obvious, and would much benefit from being documented as an example) but it looks like it fits the bill. Can't wait for a release!

PR #139, opened 2024-11-27
RFC: Pipeline Identity Tokens

dbaumgarten:
[Rendered](https://github.com/dbaumgarten/concourse-rfcs/blob/feature/pipeline-identities/139-pipeline-identity-tokens/proposal.md)


stuartpurgavie:
Would likely recommend against making RSA the default, seems to be a trend of moving towards EC25519 since RS256 uses PKCS#1v1.5 which is the vector for the Bleichenbacher style of attack, not sure increasing the bit size avoids the issues exploited in this style of attack. But should likely be default only, if the implementer is confident they have a closed ecosystem they should be able to specify other signing protocols.

dbaumgarten:
I mainly thought of RS256 because I think it has the widest support out there.
[RFC7518](https://www.rfc-editor.org/rfc/rfc7518#section-3) lists both RS256 and ES256 as recommended, so compatibility should be similar I guess.

Not sure if RS256 is actually affected by Bleichenbacher attacks. From reading into it it seems to mostly affect encryption via RSA and I couldn't find a source that says RS256 is affected.

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,133 @@
...
+    platform: linux
+    run:
+    path: bash
+    dir: idp-servicebroker
+    args:
+    - -ceux
+    - aws sts assume-role-with-web-identity --d
+      --provider-id "<ARN of the Identity Provider of Step 1>" \
+      --role-arn "<ARN of the role to be assumed>" \
+      --web-identity-token (( idtoken ))
```
dbaumgarten, line 67: I have thought a little more about this part.
The simples and cleanest approach would probably be to implement this as a [var_source](https://concourse-ci.org/vars.html#var-sources).

So it would the look like this:
```
var_sources:
- name: idtoken
  type: idtoken
  config:
    aud: ["some-audience", "some-other"]
    ttl: 15m
```

This would allow pipeline-authors to influence the generated token (in a controlled manner).
They can for example choose for which audiences a token is itended or how long it should stay valid.

The drawback is, that (as of now) it would be impossible to include anything more specific then team+pipeline in the token (so no job, task etc.) as these information is currently unavailble to a credenital-provider.

dbaumgarten:
I came up with some code to showcase how this feature could work: https://github.com/concourse/concourse/pull/9035

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,133 @@
+* RFC PR: [concourse/rfcs#139](https://github.com/concourse/rfcs/pull/139)
+
+# Summary
+
+Pipelines should receive signed JWTs ([RFC7519](https://datatracker.ietf.org/doc/html/rfc7519)) from Concourse that contain information about them (team, pipeline-name etc.).
+They could the send this JWTs to outside services to authenticate using their identity as "Concourse-Pipeline X"
```
marco-m-pix4d, line 6: the -> then

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,133 @@
...
+Vault has support for [authentication via JWT](https://developer.hashicorp.com/vault/docs/auth/jwt).
+It works similarly to AWS. You tell vault an URL to the Issuer of the JWT (your concourse instance) and configure what values you expect in the token (for example the token must be issued to a pipeline of the main team). You can then configure a Vault-ACL and even use claims from the token in the ACL. Your ACL could for example allow access to secrets stored in ```/concourse/<team>/<pipeline>``` to any holder of such a JWT issued by your concourse.
+
+Detailed usage-instructions for vault can follow if required.
+
+# Proposal
+Implementation is split into different phases, that stack onto each other. We could implement the first few and expant the implementation step by step.
+
+## Phase 1
+- When Concourse boots for the first time it creates a signature keypar and stores it into the DB
```
marco-m-pix4d, line 84: keypar -> keypair

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+# Motivation
+Often pipelines have to interact with outside services to do stuff. For example deploy something to AWS.
+As of now you would need to create static credentials for these outside services and place them into concourse's secrets-management (for example inside vault).
+
+However having static (long lived) credentials for something that is critical (like a prod account on AWS) is not state of the art for authentication.
+It would be much better if code running in a pipeline could somehow prove it's identity to the outside service. The outside service could then be configured to grant permissions to a specific pipeline.
+
+Lot's if other services already implement something like this. One well knwon example of this are [Kubernetes's Service Accounts](https://kubernetes.io/docs/concepts/security/service-accounts/#authenticating-credentials). Kubernetes mounts a signed JWT into the pod and the pod can then use this token to authenticate with Kubernetes itself or with any other service that has a trust-relationship.
```
aliculPix4D, line 16: typos: 
1. `Lot's if other services` ---> `Lots of other services`
2. `One well knwon example` ---> `One well known example`

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Usage with AWS
+For example a Pipeline could use AWS's [AssumeRoleWithWebIdentity API-Call](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html) to authenticate with AWS using it's concourse-token and do stuff in AWS. It is even [diretly supported by the AWS CLI](https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role-with-web-identity.html)
```
aliculPix4D, line 19: typo:
`diretly supported by the AWS CLI` --> `directly supported by the AWS CLI`

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+    ]
+}
+```
+This trust-policy allows everyone to assume this role via the AssumeRoleWithWebIdentity API-Call, as long as he has a JWT, signed by your Concourse, with the sub-value of "main/deploy-to-aws".
+
+And conveniently Concourse will create exactly such a token and supply it to (and only to) the pipeline "deploy-to-aws" in the "main" team.
+
+When code inside a pipeline performs the AssumeRoleWithWebIdentity API-Call, AWS will check the provided token for expiry, query concourse to obtain the correct signature-verification key and use it to check the JWT's signature. It will then compare the aud-claim of the token with the one specified in the Role's trust policy. If everything checks out, AWS will return temporary AWS-Credentials that the pipeline can then use to perfor actions in AWS.
```
aliculPix4D, line 51: typo: `to perfor actions in AWS` ---> `to perform actions in AWS`

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,133 @@
...
+## Usage with vault
+The feature would also allow pipelines to authenticate with vault. This way a pipeline could directly access vault and use all of it's features, not only the limited stuff that is provided by concourse natively.
+
+Vault has support for [authentication via JWT](https://developer.hashicorp.com/vault/docs/auth/jwt).
+It works similarly to AWS. You tell vault an URL to the Issuer of the JWT (your concourse instance) and configure what values you expect in the token (for example the token must be issued to a pipeline of the main team). You can then configure a Vault-ACL and even use claims from the token in the ACL. Your ACL could for example allow access to secrets stored in ```/concourse/<team>/<pipeline>``` to any holder of such a JWT issued by your concourse.
+
+Detailed usage-instructions for vault can follow if required.
+
+# Proposal
+Implementation is split into different phases, that stack onto each other. We could implement the first few and expant the implementation step by step.
```
aliculPix4D, line 81: expant --> expand

dbaumgarten:
@aliculPix4D @marco-m-pix4d : So, what can be done to go forward with this RFC? What would be the next steps?

marco-m-pix4d:
Hello @dbaumgarten, both myself and @aliculPix4D are only _contributors_.
Let's ask @taylorsilva if he could please chime in.

taylorsilva:
Hey folks, I don't have time to go over RFC's right now. No maintainers have reviewed this RFC yet so that'll need to happen next, which basically is me reviewing it. I'll have time to do that once I get the next release out.

dbaumgarten:
@taylorsilva I really don't want to be annoying, but now that it seems like the new release is out (yay!):
Would you have time to take a look at this PR?
I would really like to know whether it already makes sense to start working on a fully working implementation of this.

taylorsilva:
Thanks for the ping! I'm a bit deep in the ARM release stuff right now but I did want to make some time to review this stuff. I'll get to it this week as I think I've put this off long enough 😅

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,155 @@
...
+That endpoint could reject any token for a pipeline which is currently not running at all.
+
+# Open Questions
+
+1. How to call the var_source. "idtoken"?
+2. How to call the "field" for the token in the var_source. "token"?
+3. What kind of keypair is used by default? RSA-Keys with 4096 bits? We could generate multiple types of keypairs and allow pipelines to choose one via the var-source-config.
+4. What exactly to put into the sub-claim? The easiest would be "team/pipeline". But what about the job-name or instance-vars? Unfortunately instance-vars and job-names are currently not available to var-sources.
+5. How long should the default ttl be?
+6. How often to rotate the signing key?
```
taylorsilva, line 147: I do not have a good suggestion for this. Could you provide some research as to what would be a good default value here? Are there other systems that provide a similar ability like what this RFC proposes?

I found this page from MS: https://learn.microsoft.com/en-us/entra/identity-platform/signing-key-rollover

They don't give an explicit rotation timeframe but do strongly suggest that they could be frequent.

From https://learn.microsoft.com/en-us/entra/identity-platform/signing-key-rollover#general-considerations
> The time-to-live of keys in the cache should be configured to 24 hours

This feels like an "it depends" question, where the answer depends on your organizations own security posture. I'm open to any suggestions folks have.

Just to give us a starting point: rotate the key once a week and keep the previous key available for 1 day after rotation (I pulled those numbers from the air).

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,155 @@
...
+7. Can and should we include more specific information into the token (job-name/id, task, infos about the worker)?
+
```
taylorsilva, line 149:
```suggestion
8. How do pipeline identity tokens work with resources?

```

I'm raising this as a question because it wasn't mentioned at all. Since the identity token changes (TTL changes), referencing it in a resource would constantly reset the version history of the resource. This may be fine depending on how the resource behaves.

I don't want identity tokens to change any of the current behaviour of resources, but I do want us to think through how they would work for resources.

I think the `sub` claim should also be different too for resources: `team/pipeline/resource`.

taylorsilva (commented):
Thank you for your patience! I like this RFC and would love to see this added to Concourse. I appreciate the time you put into this.

I like the use of `var_sources` as the way to create the tokens. Seamlessly integrates this into existing pipeline configs.

Let's get some final answers for questions 3-8. Once we have answers to those I'll accept the RFC.

139-pipeline-identity-tokens/proposal.md
Shegox: GitHub Actions for example has a lifetime specified for it's signing key (https://token.actions.githubusercontent.com/.well-known/jwks) of 10 years:
```
        Version: 3 (0x2)
        Serial Number:
            0e:7c:26:e9:e4:48:a8:61:40:de:8f:cf:d6:80:6b:0a:1b:f1:fe:7a
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=actions.self-signed.github
        Validity
            Not Before: Jan 23 15:25:36 2024 GMT
            Not After : Jan 20 15:25:36 2034 GMT
        Subject: CN=actions.self-signed.
```

Looking at the wayback machine in the last year these keys haven't be changed: https://web.archive.org/web/20240424011427/https://token.actions.githubusercontent.com/.well-known/jwks

Google Kubernetes Engine nowadays bootstrap clusters with a [30-year lifetime]https://cloud.google.com/kubernetes-engine/docs/how-to/credential-rotation?hl=en#credential-lifetime) and a rather manual process of rotating them.

--- 

Another thing to keep in mind, if you have an "internal" `openid-configuration` endpoint it is sometimes nessecary to "hard copy" the signing keys to external platforms. 

GCP for example allows this via this UI prompt:
grafik


Kubernetes [structured authentification](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#using-authentication-configuration) via hosting the jwks under a different URL (which might be manually updated). I personally have used GCP static buckets to host the jwks for kubernets clusters and updated them manually: 
```yaml
    # The "issuer" field in the fetched discovery information must match the "issuer.url" field
    # in the AuthenticationConfiguration and will be used to validate the "iss" claim in the presented JWT.
    # This is for scenarios where the well-known and jwks endpoints are hosted at a different
    # location than the issuer (such as locally in the cluster).
    # discoveryURL must be different from url if specified and must be unique across all authenticators.
    discoveryURL: https://discovery.example.com/.well-known/openid-configuration
```

--- 

TL;DR: Long story short, I see a lot of scenarios where the quick rotation of the signing keys would make the usage harder and it seems that other platforms don't rotate them quickly either. So while, from a security perspective I find a regular rotation great, I fear this could hinder adoption.
If an automated rotation is implemented I would recommend to have a switch to turn it off and let an operator manage the 3-step rotation manually:
1. trigger creation of new keys and add them to jwks endpoint (this allows clients to distribute them)
2. start using new keys to sign things (actuall switching over, while keeping the old keys as still valid)
3. remove old keys from jwks (finally removing the old keys, if they no longer are needed)

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,155 @@
...
+## Phase 3
+To make sure tokens are as short-lived as possible we could enable online-verification of tokens. Concourse could offer a Token-Introspection-Endpoint ([RFC7662](https://datatracker.ietf.org/doc/html/rfc7662)) where external services can send tokens to for verification.
+That endpoint could reject any token for a pipeline which is currently not running at all.
+
+# Open Questions
+
+1. How to call the var_source. "idtoken"?
+2. How to call the "field" for the token in the var_source. "token"?
+3. What kind of keypair is used by default? RSA-Keys with 4096 bits? We could generate multiple types of keypairs and allow pipelines to choose one via the var-source-config.
```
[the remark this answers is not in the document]
Shegox: GtiHub Actions seems to be using only `RS256`: https://token.actions.githubusercontent.com/.well-known/jwks 

Kubernetes defaults for it's service account signing as well to `RS256` with supporting others (https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/). 

I personally think first just implementing RS256 would be fine. Support the configuration on instance level would be an okayish middleground. Selectable keys for pipelines are cool, but probably a bit overkill :)

Shegox (commented):
Really love to see this being implemented. 

Just wanted to share some of the insights I had from other tooling, which might be helpful. :)

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,155 @@
...
+Pipelines should receive signed JWTs ([RFC7519](https://datatracker.ietf.org/doc/html/rfc7519)) from Concourse that contain information about them (team, pipeline-name etc.).
+They could then send this JWTs to outside services to authenticate using their identity as "Concourse-Pipeline X"
+
+
+# Motivation
+Often pipelines have to interact with outside services to do stuff. For example deploy something to AWS.
+As of now you would need to create static credentials for these outside services and place them into concourse's secrets-management (for example inside vault).
+
+However having static (long lived) credentials for something that is critical (like a prod account on AWS) is not state of the art for authentication.
+It would be much better if code running in a pipeline could somehow prove it's identity to the outside service. The outside service could then be configured to grant permissions to a specific pipeline.
```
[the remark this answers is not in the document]
dbaumgarten: The fact that attaching trust to workers is that all pipelines inherit this trust. So a pipeline that checks a PR would theoretically have permissions to deploy to prod. Thats the whole reason why I came up with this feature.

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,155 @@
...
+## Phase 3
+To make sure tokens are as short-lived as possible we could enable online-verification of tokens. Concourse could offer a Token-Introspection-Endpoint ([RFC7662](https://datatracker.ietf.org/doc/html/rfc7662)) where external services can send tokens to for verification.
+That endpoint could reject any token for a pipeline which is currently not running at all.
+
+# Open Questions
+
+1. How to call the var_source. "idtoken"?
+2. How to call the "field" for the token in the var_source. "token"?
+3. What kind of keypair is used by default? RSA-Keys with 4096 bits? We could generate multiple types of keypairs and allow pipelines to choose one via the var-source-config.
+4. What exactly to put into the sub-claim? The easiest would be "team/pipeline". But what about the job-name or instance-vars? Unfortunately instance-vars and job-names are currently not available to var-sources.
```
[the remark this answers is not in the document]
dbaumgarten: I found out that most cloud-providers only support exact-matches on the sub-claim.
However sometimes you would want to grant access to (for example) all pipelines of a team.
In that case a sub-claim of "team" would be really helpfull.

How about letting the user choose via the var-source (in a limited way) what to put in the sub-claim.
I was thinking about something like:

```
subjectScope: team | pipeline | job | task
```

Which would result in a sub-claim of ```team```, ```team/pipeline```, ```team/pipeline/job``` or ```team/pipeline/job/task```.

I think subjectScope=pipeline would be a reasonable default.

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,155 @@
...
+5. How long should the default ttl be?
```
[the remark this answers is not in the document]
dbaumgarten: I would additionally suggest limiting the configurable ttl to at max 24h.
Otherwise someone could generate a basically unlimited valid token

139-pipeline-identity-tokens/proposal.md
```diff
@@ -0,0 +1,155 @@
...
+6. How often to rotate the signing key?
+7. Can and should we include more specific information into the token (job-name/id, task, infos about the worker)?
```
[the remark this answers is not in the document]
dbaumgarten: I agree. We should keep worker-info out of this.
Adding job (or even task) info would be great.

139-pipeline-identity-tokens/proposal.md
dbaumgarten: Most implementations should be completely fine with quick rotations. If they don't have the verification-key for a given token cached the simply query the jwks-endpoint again.

How about making the rotation-period configurable via an atc-setting and setting the default to 7 days. A rotation-period of 0 would disable automatic rotation completely.

I proposed a maximum lifetime for tokens of 24h. That would mean after 24h an outdated key can be safely removed from the jwks.

139-pipeline-identity-tokens/proposal.md
dbaumgarten: Honestly I have no idea how to handle this.
I am open for any suggestions.

dbaumgarten:
Hi,

thanks for the suggestions and answers.
I have updated the RFC accordingly.

139-pipeline-identity-tokens/proposal.md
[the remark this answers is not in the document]
Shegox: I enjoy how GitHub handles the customization of the `subjectClaim` via https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#customizing-the-subject-claims-for-an-organization-or-repository

As an option we could call the option `subjectClaims` and make it an array of included values. (I'm not really a Concourse expert, so not sure if there is an value of having just the `pipeline`-name without team as subject).

--- 

One thing we should ensure is that the seperator between the values can not be used be set by a team. E.g. can a team name include a `/`? If yes I could create a team called `my-admin-team/admin-pipeline/admin-job/admin-task` and using scope `team` impersonate the actual `admin-task` from the `my-admin-team`.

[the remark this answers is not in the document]
dbaumgarten: We really need to be super careful here to not accidentally allow impersonation.

I dont think having just the Pipeline Name as sub would be helpful. I even think it would be dangerous, as any other Team could just create a Pipeline with the same name and impersonate the first Pipeline.

dbaumgarten:
I was also thinking about renaming the fields of the var-source:
aud -> audience
ttl -> expiresIn

I think that would make things a little clearer. No need to habe overly short names there.

139-pipeline-identity-tokens/proposal.md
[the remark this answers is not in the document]
taylorsilva: Sounds like RS256 as the first and only option would work fine. Let's do that for a first implementation then.

139-pipeline-identity-tokens/proposal.md
taylorsilva: Thanks for all that info @Shegox!

---

I think your suggestion makes sense @dbaumgarten. Giving people a way to disable rotation all together sounds fine to cover the edge cases that would prompt such a configuration.

I'm good with going with your suggestion of default 7 days.

taylorsilva (changes requested):
Updates look good. Can you add a section that outlines all the config options for the `var_source`, like the `subject_claims`, please.

> I was also thinking about renaming the fields of the var-source:
aud -> audience
ttl -> expiresIn


I'd prefer the longer key names too. Nit; stick with snake-case since that's what we currently do.

dbaumgarten:
Done. Added a section with details about the var-source.
We should absolutely stick with snake_case, if thats already use everywhere else.

I also fixed a few typos here and there.

139-pipeline-identity-tokens/proposal.md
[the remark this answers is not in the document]
kcbimonte: While a pipeline today may just warn you (I need to triple check that), the goal in the future at one point was to make the following validation fail a pipeline set instead of just warn against it. 
- https://concourse-ci.org/config-basics.html#schema.identifier

Because of this, and because `/` is generally reserved for path traversal in the URL, I'd recommend against creating any team, pipeline, job, or task with a `/` in it.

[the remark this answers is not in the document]
dbaumgarten: As long as concourse does not guarantee that no team or pipeline name ever contains a "/" the var-source must actively escape such characters before putting them into the sub-claim.

Otherwise a malicious actor could create teams/pipelines with slashes to impersonate pipelines of other teams.

taylorsilva:
That's fine with me. I went back and read the RFC for JWT's: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3

Makes sense to me to provide null instead of an empty array in case it leads to the JWT being rejected by the upstream system. Doesn't read like it would, but I see what you're getting at with suggesting the default value be null/nil.

taylorsilva (approved):
Accepted! Thanks @dbaumgarten and looking forward to the implementation. Ping me in your PR if you need help or feel free to join the discord and message in the dev channel: https://concourse-ci.org/project.html

dbaumgarten:
Great!
I am quite amazed how quick and productive the discussion about this RFC was :)

I already made some more progress on the implementation today. I will ping you once it's a fully working prototype. I will probably need some help sooner or later to properly fit this feature into the codebase.

PR #140, opened 2025-09-01
RFC: Runner host mounts

tcmal:
[Rendered](https://github.com/tcmal/concourse-rfcs/blob/master/140-runner-host-mounts/proposal.md)

Discussed briefly in concourse/concourse#4241

### DIFF
--- a/140-runner-host-mounts/proposal.md
+++ b/140-runner-host-mounts/proposal.md
@@ -63,9 +63,6 @@ If validation has succeeded, each requested path is mounted into the task's cont
 
 # Open Questions
 
- - If the worker is itself running in a docker container, the mount will need to also be specified there. This will need to be made clear to users of the feature
- - Mounting directories from the host opens users up to weird permissions errors, which are often counterintuitive. For instance, group membership is not inherited from the worker user, so if the mounted directory is not world-accessible and its user/group ID doesn't match with what's specified in the container, users will get permission errors.
-
 # Answered Questions
 
 # New Implications
@@ -73,3 +70,15 @@ If validation has succeeded, each requested path is mounted into the task's cont
 Adding this feature allows users to run new, increasingly common, types of workloads using Concourse.
 Whilst it could also be used to introduce worker state, which we consider an antipattern, the intended use and limitations
 of this feature should discourage users from doing so unnecessarily.
+
+# Workers in docker
+
+If the worker is itself running in a docker container, the mount will need to also be specified there. This will need to be made clear to users of the feature.
+
+This will need to be made clear in documentation.
+
+# Permissions errors
+
+Mounting directories from the host opens users up to weird permissions errors, which are often counterintuitive. For instance, group membership is not inherited from the worker user, so if the mounted directory is not world-accessible and its user/group ID doesn't match with what's specified in the container, users will get permission errors.
+
+This will need to be made clear in documentation, and should be dealt with by operators.
--- a/140-runner-host-mounts/proposal.md
+++ b/140-runner-host-mounts/proposal.md
@@ -1,0 +1,75 @@
+# Summary
+
+This proposal introduces a mechanism through which tasks can request files, directories, or
+block devices to be mounted from the host the worker is on to the task's container.
+
+The worker must be configured to allow tasks to mount a given location, which we will make clear
+is a last-resort option when other mechanisms are unsuitable.
+
+# Motivation
+
+When performing tasks that require access to a device on the host, such as a GPU, it is required to
+make the 'files' that the kernel creates available in a task's container.
+
+For instance, to use an AMD GPU within a container, the container needs access to these paths from the host:
+
+  - `/dev/kfd` - GPU compute interface
+  - `/dev/dri` - folder containing an interface for each GPU.
+
+Currently, it is not possible to run these workloads using Concourse.
+
+# Proposal
+
+## Worker configuration option `--allowed-host-mounts`
+
+Add a new configuration option to the worker command, `--allowed-host-mounts`:
+The value should be a regular expression. Any host path that the expression fully matches will be *allowed* to be
+mounted into a task's container if it is requested.
+
+This option can be specified only once, but can use the regex OR operator `|` to match separate paths as required.
+
+## Task configuration option `host_mounts`
+
+Add a new configuration key (`host_mounts`) to the schema for `task` steps, which can be used to request paths from the host to be
+mounted into that task's container.
+
+Formally:
+
+  - `host_mounts` is a list of `host-mount-config`s, defaulting to `[]`
+  - A `host-mount-config` is either:
+    - An object, with the following keys:
+      - `host` (required, non-empty): The desired path to mount from the host
+      - `container` (optional): The mount's location in the container. Defaults to the same as `host`
+      - More options may be added down the line, as required.
+    - A string, of format:
+      - `host:container`, where `host` and `container` are strings not containing `:`, and have meanings as defined above.
+      - or, `host`, leaving `container` to default to `host` as above
+      - This is just a shorthand syntax, similar to the one Docker & Docker Compose uses.
+
+## New behaviour when spawning task containers
+
+If the task's `host_mounts` list is empty, then behaviour is unchanged.
+
+If not, then each element in the list must be validated:
+
+  - The `host` key must be fully matched by the `--allowed-host-mounts` regex.
+    - If the `--allowed-host-mounts` option was not specified, validation fails automatically.
+  - The path referred to by the `host` key must exist from the worker's perspective.
+    - Permissions, writability, etc, are not checked.
+
+If any of the elements do not validate, the task fails with an error message explaining why.
+
+If validation has succeeded, each requested path is mounted into the task's container, using the container runtime's native feature for doing so.
+
+# Open Questions
+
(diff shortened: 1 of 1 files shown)


taylorsilva:
Thanks for the RFC! I'll get around to reviewing it soon.

140-runner-host-mounts/proposal.md
```diff
@@ -0,0 +1,76 @@
...
+
+Add a new configuration option to the worker command, `--allowed-host-mounts`:
+The value should be a regular expression. Any host path that the expression fully matches will be *allowed* to be
+mounted into a task's container if it is requested.
+
+This option can be specified only once, but can use `|` to match separate paths as required.
+
+## Task configuration option `host_mounts`
+
+Add a new configuration key (`host_mounts`) to the `task-config` schema, which can be used to request paths from the host to be
```
taylorsilva, line 33: I'm debating if we should make `host_mounts` in the `task-config` or the `task-step` schema.

`task-config`
```yaml
- task: my-task
  config:
    platform: linux
    ...
    host_mounts:
      - /host/path:/container/path
```

or `task-step`
```yaml
- task: my-task
  host_mounts:
    - /host/path:/container/path
  config:
    platform: linux
    ...
```

I'm debating this because this config option feels similar to the `task.privileged` option, which is only configurable in the `task-step` schema. I know `privileged` is only configurable there because we wanted to make it easy for folks to see where in their pipelines "risky" stuff was happening.

I'm leaning towards putting this in the `task-step` schema. Wdyt?

140-runner-host-mounts/proposal.md
```diff
@@ -0,0 +1,76 @@
...
+    - Permissions, writability, etc, are not checked.
+
+If any of the elements do not validate, the task fails with an error message explaining why.
+
+If validation has succeeded, each requested path is mounted into the task's container, using the container runtime's native feature for doing so.
+
+# Open Questions
+
+ - If the worker is itself running in a docker container, the mount will need to also be specified there. This will need to be made clear to users of the feature
+ - Mounting directories from the host opens users up to weird permissions errors, which are often counterintuitive. For instance, group membership is not inherited from the worker user, so if the mounted directory is not world-accessible and its user/group ID doesn't match with what's specified in the container, users will get permission errors.
```
taylorsilva, line 67: Would permission issues be mitigated if users ran their tasks as `privileged: true`? Bit of a hack, but can probably suggest that in the docs.

140-runner-host-mounts/proposal.md
```diff
@@ -0,0 +1,76 @@
...
+ - As it stands, no consideration to host mounts is proposed during scheduling. On heterogenous clusters, worker tags can be used for this, but is this good enough?
```
taylorsilva, line 68: 👍

taylorsilva (commented):
Thanks so much for this RFC. No major blockers here, just some nit-pick stuff. Interested in your thoughts. I'll start trying to share the RFC as well to see if other community members have thoughts about it.

140-runner-host-mounts/proposal.md
tcmal: I guess it makes sense to put it in the same place as privileged

140-runner-host-mounts/proposal.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Proposal
+
+## Worker configuration option `--allowed-host-mounts`
+
+Add a new configuration option to the worker command, `--allowed-host-mounts`:
+The value should be a regular expression. Any host path that the expression fully matches will be *allowed* to be
+mounted into a task's container if it is requested.
+
+This option can be specified only once, but can use the regex OR operator `|` to match separate paths as required.
```
Rukenshia, line 29: Might be better to use a list of regexes, otherwise it might get ugly if you need many different paths.

analytically:
I'm happy to help implement this @tcmal

140-runner-host-mounts/proposal.md
```diff
@@ -0,0 +1,75 @@
...
+  - `host_mounts` is a list of `host-mount-config`s, defaulting to `[]`
+  - A `host-mount-config` is either:
+    - An object, with the following keys:
+      - `host` (required, non-empty): The desired path to mount from the host
+      - `container` (optional): The mount's location in the container. Defaults to the same as `host`
+      - More options may be added down the line, as required.
+    - A string, of format:
+      - `host:container`, where `host` and `container` are strings not containing `:`, and have meanings as defined above.
+      - or, `host`, leaving `container` to default to `host` as above
+      - This is just a shorthand syntax, similar to the one Docker & Docker Compose uses.
```
taylorsilva, line 47: Could we add an example of using `host_mounts` as well please? Similar to what you showed in this comment: https://github.com/concourse/concourse/issues/4241#issuecomment-3188638075

140-runner-host-mounts/proposal.md
```diff
@@ -0,0 +1,75 @@
...
+
+If validation has succeeded, each requested path is mounted into the task's container, using the container runtime's native feature for doing so.
+
+# Open Questions
+
+ - If the worker is itself running in a docker container, the mount will need to also be specified there. This will need to be made clear to users of the feature
+ - Mounting directories from the host opens users up to weird permissions errors, which are often counterintuitive. For instance, group membership is not inherited from the worker user, so if the mounted directory is not world-accessible and its user/group ID doesn't match with what's specified in the container, users will get permission errors.
+
+# Answered Questions
+
```
taylorsilva, line 67:
```suggestion
# Answered Questions

* > Which container runtimes (guardian and containerd) will support this feature? Can we support it in both runtimes?
  
  Looks like we can have this feature added to both runtimes. The garden client is the interface we have between both the guardian and containerd runtimes. We'll need to do the plumbing work ourselves for the containerd runtime, but we'll get this for free with the guardian runtime.
https://github.com/cloudfoundry/garden/blob/f0775181931df979cbeee43604327bb207ae0c9d/client.go#L190-L208

```

Adding and answering my own question.

taylorsilva (commented):
@tcmal Sorry for the long silence from my end. I want to put momentum behind this RFC again.

I've read it again and noticed nothing mentioned about read/write permissions of the mounts. Who has control over that? Do operators get to decide if the mounts are rw/ro or is that up to the task writer to decide?

My instinct tells me we should probably let the operators decide between between rw/ro permissions for each mount. Maybe we could allow task writers to downgrade to ro for their mounts even if the operator allows rw for the given mount. What are your thoughts on this?

140-runner-host-mounts/proposal.md
```diff
@@ -0,0 +1,75 @@
...
+  - The path referred to by the `host` key must exist from the worker's perspective.
+    - Permissions, writability, etc, are not checked.
+
+If any of the elements do not validate, the task fails with an error message explaining why.
+
+If validation has succeeded, each requested path is mounted into the task's container, using the container runtime's native feature for doing so.
+
+# Open Questions
+
+ - If the worker is itself running in a docker container, the mount will need to also be specified there. This will need to be made clear to users of the feature
```
taylorsilva: Would be the same situation for users running Concourse on k8s, which there are a lot of.

Either way, the answer here is to document it. Can move this to the Answered section.

140-runner-host-mounts/proposal.md
```diff
@@ -0,0 +1,75 @@
...
+ - Mounting directories from the host opens users up to weird permissions errors, which are often counterintuitive. For instance, group membership is not inherited from the worker user, so if the mounted directory is not world-accessible and its user/group ID doesn't match with what's specified in the container, users will get permission errors.
```
taylorsilva: Let's move this to the Answered section as well, with the answer being that we document that the onus is on operators to ensure their host mounts have the correct permissions and suggesting some useful defaults.

kcbimonte:
> @tcmal Sorry for the long silence from my end. I want to put momentum behind this RFC again.
> 
> 
> 
> I've read it again and noticed nothing mentioned about read/write permissions of the mounts. Who has control over that? Do operators get to decide if the mounts are rw/ro or is that up to the task writer to decide?
> 
> 
> 
> My instinct tells me we should probably let the operators decide between between rw/ro permissions for each mount. Maybe we could allow task writers to downgrade to ro for their mounts even if the operator allows rw for the given mount. What are your thoughts on this?

If thinking rw permissions, may want to reframe this RFC a bit. At first I thought of this as ro like mounting self signed certificates but to introduce write could get weird. 

What if instead we "extend" the get and put prototypes to make it clear we are getting files from the host and, if allowed, putting files onto the host. That opens up Concourse, for better or for worse, to use worker local storage as job gates if we use file checksums as the version. 

Obviously this is an entirely different direction than just mounting a folder, but just an initial thought when reading the proposal to allow rw permissions.

tcmal:
> I've read it again and noticed nothing mentioned about read/write permissions of the mounts. Who has control over that? Do operators get to decide if the mounts are rw/ro or is that up to the task writer to decide?

As this was written with mainly mounting GPUs in mind, it assumes that all mounts will be RW. It could probably be extended to support specifying read-only mounts as separate keys. Or, a `host_mounts` key could be RO and `rw_host_mounts` could be RW to make it more explicit.

> What if instead we "extend" the get and put prototypes to make it clear we are getting files from the host and, if allowed, putting files onto the host. That opens up Concourse, for better or for worse, to use worker local storage as job gates if we use file checksums as the version.

This is interesting, but I think it doesn't solve the problem that this RFC is trying to.

PR #141, opened 2025-09-25
Create RFC for health endpoint

Kump3r:
[Rendered](https://github.com/Kump3r/rfcs/blob/master/141-health-endpoint/proposal.md)
Previously discussed as well, but I think it makes sense to collect comments on this again, due to industry standards and an overall need for this. Should we open a discussion to collect interest/opinions of the community as well?

### DIFF
--- a/141-health-endpoint/proposal.md
+++ b/141-health-endpoint/proposal.md
@@ -1,0 +1,56 @@
+# Reference
+* Original RFC [#46](https://github.com/concourse/rfcs/pull/46)
+* Original POC [#4818](https://github.com/concourse/concourse/pull/4818)
+
+# Summary
+This proposal outlines the beginnings of support for a `health` endpoint, which has a simple backend service which monitors crucial Concourse interfaces such as the database connectivity, the worker count (of healthy workers (*should be some threshold depending on the update strategy*)), the state of the webs (ATC/TSA) and others.
+
+# Motivation
+#### Currently, Concourse does not expose a dedicated, standardized health endpoint that external systems can query to determine the system’s overall health. This creates challenges in the following areas:
+
+### 1. Monitoring & Alerting
+Operators and platform teams often integrate Concourse with monitoring systems (e.g. Prometheus, Datadog, Kubernetes liveness/readiness probes). Without a clear health endpoint, they must rely on indirect signals (such as API responses, metrics, or manual checks), which can be unreliable or difficult to standardize.
+
+### 2. Automation & Self-Healing
+Modern infrastructure frequently depends on health endpoints for automated actions like restarting unhealthy pods, removing failing nodes from load balancers, or scaling workloads. The lack of a health endpoint makes such automation harder to implement for Concourse.
+
+### 3. User Experience
+When Concourse becomes partially degraded (e.g. workers are down, ATC is unresponsive, DB is lagging), it is not immediately obvious to users or operators. A health endpoint would provide a quick, single source of truth for identifying issues.
+
+### 4. Consistency with Industry Standards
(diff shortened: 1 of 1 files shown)


141-health-endpoint/proposal.md
```diff
@@ -0,0 +1,56 @@
...
+
+# Proposal
+## API Changes
+What comes to mind is a simple **unauthenticated** HTTP endpoint (e.g. `/health`) that returns a JSON payload indicating the overall health status of the Concourse system. Could be something simple like:
+```json
+{
+  "status": "healthy/unhealthy",
+  "details": {
+    "database": "healthy/unhealthy",
+    "workers": "healthy/unhealthy",
```
radoslav-tomov, line 34: "workers" as a single entry does not convey much meaningful information IMO. A list of the status of each worker might be more useful.  Also, the semantics of the general "status" should be clarified. What is considered a healthy instance  ?

Kump3r:
I had an offline discussion with a stakeholder who also raised a valid question that can be added to the document:
"How is Concourse on K8s determining the state of the pods?"
- There are liveness and readiness probes defined in the chart, which make a http request to the `/api/v1/info` endpoint. The idea of the change would be to have a more dedicated endpoint that could build a bit on that static endpoint checking by also considering the status which can change dynamically.

Thanks for the question I will also add this to the document when a bit more questions are gathered!

taylorsilva:
Totally for this. A lot of my questions come around implementation. I think it would be nice for this RFC to define specifically what we want the Health JSON response to look like.

A Concourse web node is made up of a bunch of micro-service-ish components. We could potentially display the health of all of these components ([see components.go](https://github.com/concourse/concourse/blob/42388cf60c066c38049f34aec01db1fa3c9607ae/atc/component.go)). There may be some exceptions in that file, but most of these components are run "globally" across one of the web nodes based on workload the web node is handling. They're load-balanced!

There are some services on the web node that are not load-balanced, like the TSA and API. Those are always running on all web nodes.

A detailed health response could look something like this, which I think would accurately describe the entire Concourse cluster:

```json
{
    "status": "...",
    "workers": {
        "worker-1": {
            "baggageclaim": "...",
            "garden": "..."
        }
        ...
    },
    "web-nodes": {
        "web-1": {
            "api": "...",
            "tsa": "..."
        }
        ...
    },
    "global-components": {
        "garbage-collection": "...",
        "lidar": "...",
        "secret-management": "...",
        "scheduler": "..."
        ...
    }
}
```

I wouldn't expect an initial PR to fully implement all of that though. I think this RFC could clearly define what we want the end goal to look like and then slowly work towards it through multiple PR's.

Kump3r:
I agree, I really like the idea and I am all for having an easy to reach status board for all of the components. One of the key questions that come to mind is when the overall status should change to not healthy, as albeit each component having its share of work to be done, if they flap, or are unstable it shouldn’t mean the instance is not operational, but rather somewhat degraded. So more, or less building on what you wrote, it would be great before closing the RFC to have the json response and the conditions that are hard requirement for a healthy instance figured out. Thanks to all for the feedback, I like the overall direction of the discussions here. Once we have q couple of more comments, I will add all the discussions to the document.

Kump3r:
Hello @concourse/maintainers, I am returning to this topic as I was quite overwhelmed the last months, sorry for that 🤷. Wanted to hear your thoughts on this RFC and understand what are the next steps we could take?  Do you see any blockers in such an implementation, as it will require quite the investment and I want to make sure we are all on the same page, before hacking away at a POC. Many thanks for all of the responses this RFC gather, everything was very useful to me and you opened my eyes for a lot of details that I would have surely missed.

taylorsilva:
I'll look over this again. Just did a quick scan and have a few questions and things I want to research. I think we can iterate a little bit more on the RFC and clarify a few more things.

taylorsilva (approved):
I don't think there's anything of concern that we need to address with the RFC. I'm going to merge since our intention is to add such a feature into Concourse. How it actually gets implemented will likely drift from this RFC as the PR's get worked on.

Thanks for your work on this Kalin!

